PR #181, opened 2014-12-09
Remove basePath from resource list

gunnarlium:
basePath is not a valid element in the resource list, and it causes
swagger-ui to load the wrong paths from the backend, in effect breaking
the entire service.

See this for a related issue: https://github.com/swagger-api/swagger-ui/issues/490


bfanger:
Thanks, merged in the 0.9 branch

PR #194, opened 2015-02-10
Fix partial_id

nunovieira:
Changed gh-pages to show that the property to make an annotation become a partial is *partial*, not *partial_id*


bfanger:
Thanks!

PR #197, opened 2015-03-04
added Patch method Definition

Arilas:
Added annotation for Patch method


bfanger:
Thanks!
I always forget about PATCH request ;-)
Merged into 2.x

PR #210, opened 2015-04-25
Style fixes

mpajunen:
While browsing the code I fixed a few issues, mostly doc blocks and some typos but also a couple of missing / extra use statements. Some fixes may be stylistic and if they don't fit I can remove them.


bfanger:
Nice work! Thanks.

Issue #175, opened 2014-11-05
PSR-2 Compliance


adammbalogh:
Do you plan to make this pakcage PSR-2 Compliant?

bfanger:
[phpcs](https://github.com/squizlabs/PHP_CodeSniffer) still has some complaints:

> Constants must be uppercase; expected 'SWAGGER\UNDEFINED' but found 'Swagger\UNDEFINED'

The constant *is* uppercase, it seams phpcs doesn't take namespaces into account.
> Line exceeds 120 characters; contains ... characters

I'm okey with long lines.
> Property name "..." should not be prefixed with an underscore to indicate visibility

Properties with an underscore are used to avoid naming collisions between swagger-php and the swagger output/spec.

But the code is now mostly PSR-2 compliant, thanks to @mpajunen and [php-cs-fixer](http://cs.sensiolabs.org).

PR #211, opened 2015-04-26
Change coding style to match PSR-2

mpajunen:
PSR-2 compliance is discussed in #175. I thought I'd run [PHP CS Fixer](https://github.com/fabpot/PHP-CS-Fixer)  (`php-cs-fixer fix --level=psr2`) and this is the result. Of course this won't change everything to match PSR-2, but it could be a start.

In addition, tests now use PSR-4 autoloading directly as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,5 +34,10 @@
     },
     "require-dev": {
         "zendframework/zend-form": "~2.3"
+    },
+    "autoload-dev": {
+        "psr-4": {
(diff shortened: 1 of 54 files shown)


bfanger:
Thanks, [php-cs-fixer](http://cs.sensiolabs.org) is nice.
I've update my [Netbeans IDE](http://netbeans.org/) formatting rules to match psr2

Issue #216, opened 2015-05-13
Swagger 2.0 - Scan isn't actually excluding any directories passed into "exclude"


xiehan:
I'm doing the following:

```php
        \Swagger\scan('/srv/', [
            '/srv/identity/v1/',
            '/srv/listening/v1/',
            '/srv/producer/'
        ]);
```

yet I'm getting this error:

```
Unexpected field "paramType" for @SWG\Parameter(name="title"), expecting "ref", "parameter", "name", "in", "description", "required", "schema", "type", "format", "items", "collectionFormat", "default", "maximum", "exclusiveMaximum", "minimum", "exclusiveMinimum", "maxLength", "minLength", "pattern", "maxItems", "minItems", "uniqueItems", "enum", "multipleOf", "x" in \ProducerSearchCallback->searchPodcasts() in /srv/producer/callback/ProducerSearchCallback.inc on line 14
```

A bit of debugging has yielded that the `scan()` method is indeed failing to ignore anything passed into `exclude`. I think it's due to an incorrect usage of Symfony's Finder.

(Note that I can probably submit a PR to fix this today, but I figured I'd open a ticket first, just in case I don't manage to make any progress.)

bfanger:
I'm currious what you're supposed to pass to `Finder->exclude()`. 
let me know what you find.

xiehan:
@bfanger It's definitely lacking in documentation so it took me a while to track down the problem...

Here's what I managed to figure out:

1. Anything passed to `exclude()` needs to be a relative URL (to whatever was passed to `in()`). In other words, in my example, I needed to strip out `/srv/` from all of the paths in the `exclude` array.
2. Directories passed to `exclude()` cannot end in a trailing slash. Considering ending in a trailing slash is fine for `in()` and this behavior isn't documented anywhere, this can probably be considered a bug in Finder, but I'm also not sure it's truly worth opening a ticket with them about. (Let me know if you think this is worth doing though.)
3. You can only pass in directories to `exclude()`, not single files. After many trials and failures, I figured out that you have to use `notPath()` for single files. Again, `notPath()` needs to be relative to whatever got passed into `in()`.

The question now is: should we have Swagger-PHP try to handle these situations more gracefully (by eg. forcing exclude URLs to be relative to the director(y/ies) passed in and stripping trailing slashes) or do we just expect users to pass paths to `exclude()` in the format Finder seems to expect (by adding documentation rather than code)?

[...]

PR #217, opened 2015-05-14
Bare minimum fix for #216

xiehan:
This is the bare minimum needed to get the 'exclude' param at least somewhat working as it should; however, it puts the burden on the user to format the input to `exclude` exactly right (using only relative paths, no trailing slashes, etc).

The main change is to use `notPath` instead of `exclude` because it turns out `notPath` can handle either files or directories (which means you also don't need to do an `is_file` check in order to figure out how to handle the exclude/s); however `notPath` can't take an array, only a string, so a little extra code was needed.

See my next PR for a more user-friendly fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /vendor
 /composer.lock
+/.idea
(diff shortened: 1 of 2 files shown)


bfanger:
I used #218 ;)

PR #218, opened 2015-05-14
More user-friendly fix for #216

xiehan:
This allows either absolute or relative paths to be passed in to `exclude`; if an absolute path is passed in, it'll turn it into a relative path based on what was passed in to `directory` (which can also be an array). It'll also ensure leading and trailing slashes are trimmed when doing this input processing. This means more work on Swagger's side, but the user's experience is better because they don't need to worry about formatting the input to `exclude` exactly right to conform with Finder's undocumented expectations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /vendor
 /composer.lock
+/.idea
--- a/src/Util.php
+++ b/src/Util.php
@@ -1,0 +1,62 @@
+<?php
+
+/**
+ * @license Apache 2.0
+ */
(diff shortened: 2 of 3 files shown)


bfanger:
Excellent work. I like the Util "functions that don't neatly fit anywhere else" class.

I've merged/rebased/tweaked this pull request into ee70610faa04dea27ed2955a73cfb67c5d6383b8

I'm a bit disappointed with Symfony Finder, thanks for making it work as expected.

PR #226, opened 2015-06-30
api.php

vanbrabantf:
fixed the comments


bfanger:
I agree your version looks better, sadly doctrine/annotations doesn't strip out the leading " * " causing it to appear in generated output, which causes the unittests to fail.

File a bug to doctrine.

PR #228, opened 2015-07-08
Fixed response annotation example in readme

dmyers:
I grabbed the new release, but the readme had an outdated example. I was getting this error before the change:

```php
( ! ) Notice: Unexpected field "name" for @SWG\Response(), expecting "ref", "response", "description", "schema", "headers", "examples", "x" in 
```

Do the docs need to be updated too? http://zircote.com/swagger-php

What do you think @bfanger?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -34,7 +34,7 @@ Add annotations to your php files.
 /**
  * @SWG\Get(
  *     path="/api/resource.json",
- *     @SWG\Response(name="200", description="An example resource")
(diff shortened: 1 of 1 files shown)


bfanger:
Thanks!

The gh-pages (http://zircote.com/swagger-php) should indeed get a "These docs are for swagger-php 1.x, for 2.x the docs are located [here](https://github.com/zircote/swagger-php/tree/master/docs)" or get an update.

Contributions are welcome.

PR #233, opened 2015-08-12
Fixing missing classes in definitions when declared as "final class".

alagafonov:


bfanger:
Thanks! merged as 51dcb53

PR #236, opened 2015-08-15
Fix tiny typo

travisdoherty:


bfanger:
Good find, but the typo is (also) in the target http://petstore.swagger.io/v2/swagger.json
File the issue there.

PR #244, opened 2015-09-16
Adds support for wrapping non standard annotations in a CustomAnnotation

DerManoMann:
This change allows users to write custom handler to convert non swagger annotations to valid swagger annotations.

The implementation is left ot the user and is nothing swagger itself needs to be concerned about.

Motivation
------------
I am using Silex with annotations (https://github.com/danadesrosiers/silex-annotation-provider) and while the Silex annotations do not exactly match swagger it would be nice to be able to re-use them for swagger instead of mostly duplicating annotations.
Furthermore, I do have some custom annotations that do not really would work out of the box.

In more detail
-----------------
* A user would need to provide an implementation of the new interface Swagger\CustomAnnotationHandlerInterface
This interface defines a method to 'migrate' custom annotations to swagger annotations (typicalle Operation subtypes)
* Custom (whitelisted) annotations are wrapped in a CustomAnnotation object that is, in fact, a swagger annotation. The key point is that all other processing still happens, ie there is a context, etc.
* Custom and swagger annotations can be mixed.


DerManoMann:
Bump: Any chance of getting some feedback on this at all?

bfanger:
Is it possible to create a Processor that migrates the Silex annotations?

I'd rather not intruduce another step in the swagger-php pipeline, but i'm not sure if the Analysis exposes enough information.

This silexprocessor should run before the other processors, which you can do with an array_unshift to Analysis::processors()

DerManoMann:
Good point - I'll have a look. Thanks.

DerManoMann:
Hmm, had a quick look and it seems the processors are applied only at the end of all analysis/parsing.
That means if custom annotations are left alone they won't have a context. However, that is the key thing I need as otherwise there is no way to properly keep track of class annotations, IMO.

bfanger:
Of course, i'll see if I can attach the context to non-standard annotations.

DerManoMann:
Checked your changes on master and that is good enough for me - thanks.

I ended up morphing my custom handler into extending Analysis and hooking into addAnnotation().
I prefer to do the conversion as early as possible to avoid missing out any other processing that might happen. Also, that way I am sure that nesting and stuff still works :)

PR #246, opened 2015-09-18
Changed annotation docblocks

dmyers:
What do you think @bfanger?


dmyers:
What do you think @bfanger?

bfanger:
Thanks!, merged in b0cf646758ce7d6807678107ce95c4573abefb0e

PR #250, opened 2015-09-25
Fixed a bug with inherited properties being null

thekonz:
If a class inherits from another class without docblocks, swagger-php throws an error, because it is trying to loop through null.

I encountered this problem today. My models inherited from a class without a docblock.
When i tried to run `Swagger\scan('/path/to/app')`, it greeted me with `Invalid argument supplied for foreach()`. Not having DocBlocks in a class is certainly not the best thing to do, but it should not stop me from seeing the swagger generated docs.

I included a unit test with fixtures. If you comment out the `if` statement in `InheritProperties`, you will see, that the unit test fails with said error.


lucasmezencio:
:+1:

ovr:
:+1: 
ping @zircote

bfanger:
Thanks!

PR #251, opened 2015-10-01
Fix invalid foreach properties

lucasmezencio:


lucasmezencio:
There is another PR that fix this. :)

Issue #253, opened 2015-10-11
RFC Proposal: remove format restrictions


bravo-kernel:
Please review before I create a PR.

Quoted from the [2.0 specification](https://github.com/swagger-api/swagger-spec/blob/a9e982aff2400a934e429a8ed7b6270dff5d8694/versions/2.0.md#data-types):

> Primitives have an optional modifier property format. Swagger uses several known formats to more finely define the data type being used. However, the format property is an open string-valued property, and can have any value to support documentation needs. Formats such as "email", "uuid", etc., can be used even though they are not defined by this specification

Currently the lib is throwing notices when using non-swagger-predefined-formats:

> Notice (1024): @SWG\Parameter(name="id",in="path")->format "uuid" is invalid, expecting "int32", "int64", "float", "double", "byte", "date", "date-time" in \App\Controller\SwaggerController->index() in /home/vagrant/projects/my.cool.app/src/Controller/SwaggerController.php on line 30 [ROOT/vendor/zircote/swagger-php/src/Logger.php, line 38]

Caused by an `in_array` check against https://github.com/zircote/swagger-php/blob/master/src/Annotations/Parameter.php#L172

My suggestion would be dropping that whole array now that formats (seem to be) non restricted anyways, would you agree?

bfanger:
Agreed, thanks!

bravo-kernel:
Awesome, thanks for updating. Keep up the good work!

PR #255, opened 2015-10-23
Allow free parameter formats

bravo-kernel:
This is a follow-up to #253 and #3236b99. Without it the same error is still thrown.


bravo-kernel:
Damn, was a bit too quick there, what was I thinking? 

The fix seems valid though, do you want me to update the tests?

PR #292, opened 2016-02-13
Prepares for release 2.0.6 (with workflow question)

bravo-kernel:
Mr. @bfanger, can you confirm that I am following the correct release workflow before I start shipping 2.0.6?

1. `VERSION.txt` on master always contains the next release version suffixed with `-dev` (e.g. 2.0.6-dev)
2. Before releasing:
    - the version in `VERSION.txt` file is stripped of `-dev` suffix (so it becomes e.g. 2.0.6)
    - `Changelog.md` is updated along with refs to the PRs
3. Release is then made against the master branch
4. After the release has been created:
    - update version in `VERSION.txt` on master (to e.g. 2.0.7-dev in this example)


bfanger:
Yep, thats pretty much it. although i didn't release from the master branch, but from a release branch that is one commit ahead of origin/master (with the version change and the changelog copied into the commit message) and only pushed that commit as a tag.

bfanger:
Maybe we could detect the installed version from the composer.lock file.

bravo-kernel:
I am open for any workflow that will prevent me from f*** up so please let me know how you want to have it. My way is a bit foolproof but quite noob as well.

bfanger:
I haven't encountered an issue so far where the version of swagger-php was unclear.
So simply leaving it at 2.0.6 in master instead of a 2.0.7-dev is fine by me.

bravo-kernel:
OK, so may I suggest... releasing 2.0.6 the manual way (as described in this PR) and take the time to come up with a more automated release cycle we can discussed in an issue for future reference. I'm sure I can get some good advice on the #cakephp channel.

bfanger:
:+1:

bravo-kernel:
:white_check_mark: and.. ty for reflecting

PR #293, opened 2016-03-03
Add augmentation for operations

mjanser:
This will extract the summary and description for operations from the
phpdoc method description.

### DIFF
--- a/tests/UtilTest.php
+++ b/tests/UtilTest.php
@@ -13,7 +13,7 @@ class UtilTest extends SwaggerTestCase
 
     public function testExclude()
     {
-        $swagger = \Swagger\scan(__DIR__ . '/Fixtures', ['exclude' => ['Customer.php', 'UsingRefs.php', 'GrandParent.php']]);
+        $swagger = \Swagger\scan(__DIR__ . '/Fixtures', ['exclude' => ['Customer.php', 'UsingRefs.php', 'UsingPhpDoc.php', 'GrandParent.php']]);
         $this->assertSame('Fixture for ParserTest', $swagger->info->title, 'No errors about duplicate @SWG\Info() annotations');
     }
 }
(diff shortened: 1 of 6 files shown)


tests/UtilTest.php
```diff
@@ -13,7 +13,7 @@ class UtilTest extends SwaggerTestCase
 
     public function testExclude()
     {
-        $swagger = \Swagger\scan(__DIR__ . '/Fixtures', ['exclude' => ['Customer.php', 'UsingRefs.php', 'GrandParent.php']]);
+        $swagger = \Swagger\scan(__DIR__ . '/Fixtures', ['exclude' => ['Customer.php', 'UsingRefs.php', 'UsingPhpDoc', 'GrandParent.php']]);
```
ChristianRiesen: missing a .php? :)

mjanser: hm, true :) it worked anyway but I fixed it now.

mjanser:
Does anything speak against merging this?

bfanger:
I'd like it if the splitting of the description and summary was moved to the AugmentOperations or Util class, but didn't make the time to do that yet.

mjanser:
That's a good idea, I moved the splitting of the description to AugmentOperations. What do you think?

bfanger:
Merged as 2eec9f1f0b582a7ae30e7f3564fde3c40767512e i've updated Context->extractDescription() to include linebreaks and updated AugmentOperations accordingly, i've also made the blankline between a summary and a description a requirement.

Let me know if you disagree with the changes i've made. and thanks the pull request!

mjanser:
Thanks for merging!

The only thing I'm unsure about your changes is if the handling of line breaks is correct now. According to [phpdoc](https://phpdoc.org/docs/latest/references/phpdoc/basic-syntax.html) the description is kind of [markdown](https://daringfireball.net/projects/markdown/syntax#p) and therefore only blank lines create new paragraphs.

bfanger:
Interresting read, even specifies the rules for the destinction between summary and description.

bfanger:
Moved summary / description logic back to the Context as it applies to all phpdoc comments not just the AugmentOperation. Also implemented the splitting by the  "line ending with a `.` " specification.

mjanser:
Cool, looks good. I'll close this pull request then.

PR #303, opened 2016-04-05
Add phpunit to the project for development purposes

ChristianRiesen:
Seems weird that it's not been there already. Makes running tests much easier and more consistent.


bfanger:
Merged as a512e61

Thanks!

PR #332, opened 2016-08-22
Napespace declaration now the first statement

Phillaf:


bfanger:
👍 Thanks, merged as  91e0f64

PR #335, opened 2016-09-05
changed @ operator as array_key_exists or isset for #331

### LINKED ISSUE #331: Do not use the silence operator `@`
You are silencing errors instead of checking for existing indexes, e.g. `@$options['analyser']`.
This is really bad practice and creates alot of error logs in our app.
Please replace those with `isset` checks or similar.

hkulekci:
I try to fix #331 

ping @bfanger


bfanger:
👍 Thanks, merged as  009dbf7

PR #345, opened 2016-10-15
Use phpdoc example tag for propery example

ismailbaskin:
https://phpdoc.org/docs/latest/references/phpdoc/tags/example.html


Nebual:
This can now be closed, as it was merged in #534

PR #350, opened 2016-12-05
Update finder options to force Swagger-PHP parse only *.php file

marcoleong:
We faced a issue where swagger-php trying to parse non-php files, that causes `Unexpected token error`. I added this patch to our version of swagger-php, to parse only files with .php extensions.

Is there any use case that swagger php needed to parse non-php files?

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -75,7 +75,7 @@ class Util
             $finder = new Finder();
             $finder->sortByName();
         }
-        $finder->files();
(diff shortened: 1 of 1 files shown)


bfanger:
Sadly, some programmers used `.inc` and maybe others, but i haven't seen that in any modern php codebase. I'm in favour of parsing only `*.php` files by default.
 
If needed those users can  configure a custom Finder object and pass that to Swagger\scan()

PR #355, opened 2016-12-23
Add php7 compatibility (rename Parent test class), turn on php7 for autotests

dbajal:
I've just fixed 2 of 3 php7cc messages and launched tests under php7.


bfanger:
Thanks!

PR #364, opened 2017-01-30
Update README.md

FallDi:
remove not necessary parentheses for require argument


bfanger:
Thanks, but I prefer the optional parentheses.

PR #380, opened 2017-03-11
Fix errors when parsing php7 anonymous class

okneloper:
When parsing my project I got some error output like this
[NOTICE] Uninitialized string offset: 1 in vendor\zircote\swagger-php\src\StaticAnalyser.php on line 93

This resulted from parsing php7 anonymous classes.

I called the new fixture file 'php7.php' so that more php7 specific syntax can be added there in future.


bfanger:
Nice! thanks.

PR #393, opened 2017-04-25
enable finder search follow link files and directories

Rivsen:
If I have many classes in different directories, I can create a directory, create files and directories link in it. Add this method, swagger-php will follow this structure as well.


bfanger:
Thanks

Rivsen:
You're welcome~
Which version will include this patch?

bfanger:
v2.0.10 and up

PR #399, opened 2017-05-05
improved tests

ivoba:
I was experimenting with the Serializer and had to test "required" property as array and also stumbled over the issue with "$ref" being not serialized, which is already fixed (https://github.com/zircote/swagger-php/commit/3e319bddbeae6cf76790c59474d081cfca0d1d03).
So i improved the test to cover both cases.


bfanger:
Thanks, merged as 28b6a03

PR #403, opened 2017-05-22
AbstractAnnotation should check for all class properties

### LINKED ISSUE #359: Problems with Typo3 flow with doctrine error
In my project I am using Typo3 Flow framework.
As soon as I use this package and add some @SWG annotations  to it,
it does not load anymore.
I get the following error:
```
User Notice: Property "_types" doesn't exist in a @SWG\Xml(), exising properties: "name", "namespace", "prefix", "attribute", "wrapped", "x", "_context", "_unmerged" in \Doctrine\Common\Annotations\DocParser->Annotation() in /docker/Packages/Libraries/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php on line 816 in /docker/Packages/Libraries/zircote/swagger-php/src/Logger.php line 38
backend_1        | Type: TYPO3\Flow\Error\Exception
backend_1        |   Code: 1
backend_1        |   File:
backend_1        | Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Error/ErrorHandler.php
backend_1        |   Line: 81
```
Looks like the combination of Doctrine and swagger-php is making trouble.
Already tried to add
```
@ignoreAnnotation
```
and 
```
AnnotationReader::addGlobalIgnoredName('SWG\Model');
...
```
Any ideas what the problem could be?

johannessteu:
Resolves #359

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -126,7 +126,7 @@ abstract class AbstractAnnotation implements JsonSerializable
 
     public function __get($property)
     {
-        $properties = get_object_vars($this);
+        $properties = get_class_vars($this);
         Logger::notice('Property "' . $property . '" doesn\'t exist in a ' . $this->identity() . ', exising properties: "' . implode('", "', array_keys($properties)) . '" in ' . $this->_context);
     }
 


bfanger:
No, it shouldn't, `get_class_vars()` includes the static properties, which are not part of the annotation instance and should be accessed with`$annotation::$propertyName` not by `$annotation->propertyName`

That is why the magic [__get](http://de.php.net/manual/en/language.oop5.overloading.php#language.oop5.overloading.members) te is invoked.

PR #424, opened 2017-08-07
Fixes for #423

### LINKED ISSUE #423: PHP notice when yielding anonymous classes implementing interface with trait in source code
I'm getting the same issue as #390 even in 2.0.10 where it's declared as fixed?

Specifically, I'm "yielding" an anonymous class that implements an interface which is honoured by a trait.

schodemeiss:
This commit attempts to fix #423 along with additional data in the unit tests to catch this case.

### DIFF
--- a/src/StaticAnalyser.php
+++ b/src/StaticAnalyser.php
@@ -101,7 +101,7 @@ class StaticAnalyser
                 }
                 $token = $this->nextToken($tokens, $parseContext);
 
-                if (is_string($token) && $token === '{') {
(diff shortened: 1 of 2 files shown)


schodemeiss:
@bfanger Any change of a merge of this?

bfanger:
Yes, thanks for your contribution.

schodemeiss:
No problem. Apologies for the nag - just can't produce any swagger docs until this was unblocked. Sorry!
Thanks for a great project.

PR #429, opened 2017-08-16
When schema type is array, items key must always exist.

kilahm:
According to the spec, when a schema has type `array`, the `items` key must always exist.  This rule was being validated for `@SWG\Items()` but not for all `@SWG\Schema()`.

* Includes a new test to cover the generic `@SWG\Schema()` case.
* Keeps the old test case for `@SWG\Items()` to ensure the log message remained the same.


bfanger:
Thanks!

PR #433, opened 2017-08-31
Accept Windows/DOS line breaks in phpdoc summary

jojonas:
The phpdoc content is statically parsed (in Context.php). During parsing, the comments are split into an array of lines. Previously, the code used explode("\n", ...) in order to split at Unix (lf) line breaks. In order to also be able to accept Windows/DOS line breaks (cr + lf), explode is replaced by splitting with preg_split on either \n or \r\n.

### DIFF
--- a/src/Context.php
+++ b/src/Context.php
@@ -171,7 +171,7 @@ class Context
         if (!$content) {
             return null;
         }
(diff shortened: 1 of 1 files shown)


bfanger:
Thanks!

Issue #419, opened 2017-07-26
Error with doctrine/annotations (v1.5)


CesarCasanova:
New Version of doctrine/annotation (1.5) is generating some problems with the json generation of this library:
`EXCEPTION] Return value of Doctrine\Common\Annotations\AnnotationRegistry::registerLoader() must be an instance of Doctrine\Common\Annotations\void, none returned in /apps/neoauto.concessionaire.com/app/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationRegistry.php on line 117`

I suggest use the previous version of that library in your composer.json: 1.4; I tried with that version and this error does not occur

bfanger:
I'm not experiencing this issue on `PHP 7.1.7` on macOS 10.12.6

https://travis-ci.org/zircote/swagger-php/builds/257931704 also looks good, are you also running composer with `--ignore-platform-reqs` ?

jaccborn:
hello, guys , i have the same question when i  run  composer update, the exception:
Type error: Return value of Doctrine\Common\Annotations\AnnotationRegistry::registerLoader() must be an instance of Doctrine\Common\Annotations\void, none returned
I guess it's because the upgrade causes the problem。and i run composer require doctrine/annotations:1.4   the problem is soloved, thanks.

fgreinus:
Maybe forcing a specific version of the doctrine/annotation would be better here. Because when you take a look at any statistics on which versions of php are used, php7.1 is not heavily adopted (for now).

mpajunen:
Forcing a specific version is not a great fix, as it would actually prevent anyone from using newer version of Annotations with swagger-php. As has been mentioned, that by itself is no problem.

There are at least two ways to prevent the problem in your project:

1. Configure your PHP version properly and don't use `--ignore-platform-reqs`.
1. Use a specific version of Annotations in your project.

PR #437, opened 2017-09-07
fixate used doctrine/annotations version to 1.4.* to avoid requiring …

fgreinus:
…php 7.1

addresses #419

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     },
     "require": {
         "php": ">=5.6",
-        "doctrine/annotations": "*",
+        "doctrine/annotations": "1.4.*",
         "symfony/finder": "*"
     },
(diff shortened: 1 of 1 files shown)


bfanger:
Thanks for your effort, but I won't merge this pul request.

Composer is smart enough to install packages based on a specified version range and php version.

Version pinning in a project composer.json is good practice, but in a library we must be more liberal. Especially with a package that other packages also depend on, this is to make sure composer can find a suitable version.

When users are forcing composer to make mistakes (using `--ignore-platform-reqs`) they are also responsible for fixing version incompatibilities. 

In this case an incompatible sub-package: doctrine/annotations, which is easily fixed by specifying the 1.4.* in composer.json **of the project**

PR #439, opened 2017-09-19
Does not work with symfony/finder lower than 2.2 version

svycka:
```
1) SwaggerMiddlewareTest\GeneratorTest::testCanGenerate
LogicException: You must call the in() method before iterating over a Finder.

/home/svycka/Documents/projects/github/swagger-middleware/vendor/symfony/finder/Symfony/Component/Finder/Finder.php:518
/home/svycka/Documents/projects/github/swagger-middleware/vendor/zircote/swagger-php/src/functions.php:39
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "require": {
         "php": ">=5.6",
         "doctrine/annotations": "*",
-        "symfony/finder": "*"
+        "symfony/finder": "^2.2 | ^3.0"
     },
     "autoload": {
         "psr-4": {


bfanger:
Instead of forcing a version, could you make it work in older versions for `symfony/finder` ?

svycka:
I don't think it's worth the effort also some versions can't be fixed like version 2.0.4 https://github.com/symfony/finder/blob/v2.0.4/composer.json there is no autoloading config in composer.json and that's impossible to workaround at least easy way... Also, I am not sure that this is not a `symfony/finder` bug also I am not using this old version just tests with --prefer-lowest failed so...

bfanger:
Merged as 8a24664 
Thanks!

PR #450, opened 2017-10-12
Fix Schema::type phpdoc is not about parameters

Tobion:
Schema::$type is not for parameters but mostly properties. There is a separate class for parameters.


bfanger:
Merged as ebc8730
Thanks!

PR #451, opened 2017-10-13
Added field name allowEmptyValue

jamisonkirk:
Hello,

This change ads the Parameter annotation field name allowEmptyValue as this seems to be part of the swagger-php api spec.

https://bfanger.nl/swagger-explained/#parameterObject

Thanks,
Jamie

### DIFF
--- a/src/Annotations/Parameter.php
+++ b/src/Annotations/Parameter.php
@@ -71,6 +71,12 @@ class Parameter extends AbstractAnnotation
     public $format;
 
     /**
+     * Sets the ability to pass empty-valued parameters. This is valid only for either query or formData parameters and allows you to send a parameter with a name only or an empty value. Default value is false.
+     * @var boolean
+     */
+    public $allowEmptyValue;
+    
+    /**
      * Required if type is "array". Describes the type of items in the array.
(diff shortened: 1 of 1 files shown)


gijsstegehuis:
Really need this one.

For some reason the default value for allowEmptyValue is true.
Without this PR it's impossible to change this behavior.

bfanger:
Thanks!

gijsstegehuis:
Thanks both.
When will this be released?

bfanger:
Today! I've just released 2.0.12

gijsstegehuis:
This isn't working when used together with `@QueryParam(allowBlank=true)`.

Setting `allowBlank` to true in this annotation overrides the `allowEmptyValue` in `@SWG\Parameter()` which results in the parameter still being send as an empty string.

**Expected behavior**: `allowEmptyValue` value in `@SWG\Parameter()` should override whatever value is set in `allowBlank` of `@QueryParam`.

bfanger:
@gijsstegehuis swagger-php doesn''t look at the`@QueryParam` annotation, file an issue to the package that processes the `@QueryParam`

Issue #457, opened 2017-11-14
Add annotations


hpatoio:
I'm using ReDoc as documentation tool. I would need these `vendor` extension into my swagger: https://github.com/Rebilly/ReDoc#swagger-vendor-extensions

AFAIK at the moment `swagger-php` doesn't support these properties. What's the best way to add extra annotation ?

Would you consider a PR that add these annotations ?

bfanger:
Swagger-PHP has support for vendor extensions, how to use them is documented in the [Getting started guide](https://github.com/zircote/swagger-php/blob/master/docs/Getting-started.md)

hpatoio:
I had a look but I do not understand how add `x-logo`.

I added this to my annotations

```
 *     @SWG\Info(
 *         version="1.2.3",
 *         title="Foo API",
 *         x={"logo":"https://rebilly.github.io/ReDoc/petstore-logo.png"}
 *     )
```

and I get

```
    "info": {
        "title": "Foo API",
        "version": "1.2.3",
        "x-logo": "https://mysite.com/my_logo.svg"
    }
```

while AFAIK it should be

```
    "info": {
        "title": "Foo API",
        "version": "1.2.3",
        "x-logo": {
            "url": "https://mysite.com/my_logo.svg"
        }
    }
```

bfanger:
You'll need to define logo as an object.

```php
 *     @SWG\Info(
 *         version="1.2.3",
 *         title="Foo API",
 *         x={
 *             "logo": { 
 *                 "url": "https://rebilly.github.io/ReDoc/petstore-logo.png"
 *             }
 *         }
 *     )
```

PR #458, opened 2017-11-15
Example with array as value for custom properties

hpatoio:
Related to #457


bfanger:
> Note that Doctrine annotation supports arrays, but uses { and } instead of [ and ].

PR #469, opened 2017-12-01
Test against PHP 7.2

carusogabriel:
With the release of [`PHP 7.2`](php.net/archive/2017.php#id2017-11-30-1), would be nice to test it.


bfanger:
Thanks! merged as eb5b175916c07d7300405289d33d8a7feb55b95b

PR #472, opened 2017-12-04
3.x implement body params

mikk150:


bfanger:
I've merged another pull request that added [RequestBody](https://github.com/zircote/swagger-php/blob/3.x/src/Annotations/RequestBody.php) and more.

If that version is missing features that this PR provides, feel free to reopen.

PR #481, opened 2018-01-10
Extend OAS for petstore example 

sonrac:
Extend OAS with petstore 3.0.0 example from base example on swaggerhub


bfanger:
Nice work! thanks!

PR #483, opened 2018-01-12
Added the option: "operationid".

byjg:
I am developing a Route Generator for REST calls based on the swagger.json. 

As I am using your library to create my JSON file and you have the context where the Annotation was placed I made a small change to override the "operationid" with the Namespace, class and method of the annotation context. 

This change does not break compatibility because it is optional.

### DIFF
--- a/bin/swagger
+++ b/bin/swagger
@@ -10,6 +10,7 @@ $options = array(
     'help' => false,
     'version' => false,
     'debug' => false,
+    'operationid' => false,
 );
 $aliases = array(
     'o' => 'output',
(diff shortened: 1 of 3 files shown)


bfanger:
Nice, feature! Could you reimplement it as a processor?

I'd like to keep Analysis.php simple, the command line option would then enable the optional processor using: `Analysis::registerProcessor(new DetectOperationId())` 
or something along those lines.

byjg:
Thank you!

I did it using Processor.

byjg:
Hello, is there any comments about this PR?

bfanger:
Nope, just have to find some free time to merge it into master and the 3.x branch.
I’ll probably enable it by default on the 3.x branch.

byjg:
Good to know! As you want to merge it in the branch 3.x I created the PR #487 specific for the branch 3.x;

bfanger:
Merged as 941f0b62b7a2f2d784d5de13b146682e65449f53  

I've added some features:

- Ability to set the operationId manually in the annotation
- Ability to hide a specific operationId: `@SWG\GET(operationId=null)`.
- Also generate an operationId when not in a class and/or namespace context.
- Removed --operationid in favor of a generic --processor argument.

PR #502, opened 2018-03-07
Adding 'excludeFiles' option to exclude certain files from generation…

Acolyte:
…; works analogous to 'exclude' array and will exclude controllers where each may have SWG\Info and conflict with the others.


bfanger:
I don't think we need an additional `"excludeFiles"` option, i'd rather have the `"exclude"` option also handle the excluding of files.

Acolyte:
Agreed, I've implemented it using through existing method and closing this request.

PR #524, opened 2018-05-23
add support for use trait

gjergjsheldija:
this pull request is to add support for the following situations, where traits are used to store common properties.

```
/**
* @OAS\Schema()
*/
class className {
use traitName;
..
}

trait traitName {
/**
* @OAS\Property()
*/
var $propertyName;
}
```


gjergjsheldija:
hi @bfanger any comments on this ?

bfanger:
I did (temporary) remove the 3.x branch from GitHub (as it's now merged into master), looks like that cause the issue to be closed.

PS. There are now 2 PR addressing #522 (one for 2x and this one for 3x, i'll have to investigate which one to merge when i have some free time)

gjergjsheldija:
thank you very much for reopening it again.
just a quick clarification, which I forgot to add when first created the pull request. 
in my implementation I've treated traits as includable files, so there is no need to add ``@OAS\Schema`` at the top,that's why  they only contain ``@OAS\Property``

so everything becomes 
```php
/**
* @OAS\Schema(required={"api_name"}
*/
class demoClass {
use demoTrait;
..
}

trait demoTrait {
/**
* @OAS\Property(property="api_name")
*/
private $api_name;
..
}
```

bfanger:
Thanks! merged as  13980bf

Ps. Even though i like the stylistic changes your code formatter made, it made the PR harder to evaluate and merge.

dafrost22:
@bfanger I see this as merged, and am using the latest version at v3 but going back to what @gjergjsheldija was saying, you shouldn't need the annotation for OA\Schema at the top of the trait. However, without that annotation at the top of my trait I get the error : 
> "User Notice: Unexpected @OA\Property(), expected to be inside @OA\AdditionalProperties, @OA\Schema, @OA\JsonContent, @OA\XmlContent, @OA\Property, @OA\Items in Trait.php

I then have to add the OA\Schema annotation to the Trait, but then they are reflected as two different models in the output. Isn't the idea here to just merge the properties with the class using the trait and present them as one model? That's what my pull request for the older version (#522) was trying to do. Thanks for you time!

bfanger:
@dafrost22 Could you create a new issue? Or even better a PR with a fix 😉

PR #529, opened 2018-05-31
Merge array values when merging annotation properties

oxan:
This allows annotations that will be merged to have all the values of the property in both annotations

One usage is multiple `@Swagger` definitions which each have parameters or responses. My application is that I'm autogenerating operation annotations based on metadata from my framework, and this allows parameters in these autogenerated annotations to be merged with manually declared options for that operation.


thlassche:
@bfanger Would be nice if we could this one merged!

bfanger:
I'm concerned merging the array values is not the desired behaviour in all use-cases.

If you're already generating annotations, its a small step to also do the merging  as a post-processing step.

oxan:
There are currently only 2 call sites for `mergeProperties` in the codebase, one in the `BuildPaths` and one in the `MergeIntoSwagger` postprocessor. Looking at the array properties of the `Path` and `Swagger` annotations being merged into, merging the values seems like the desired behaviour for all of them.

bfanger:
The current implementation uses `array_merge` which is dangerous because:
> the later value for that key will overwrite the previous one
This could lead to different output depending on the order in which the annotations are scanned.

Could you update the PR so it only merges arrays without conflicts?

Issue #543, opened 2018-07-02
$ref at bottom & allOf fix


nms666:
Hi

I had a problem with the produced JSON from my schema and I made 2 changes that fixed it.
My annotation that produced wrong JSON:

    /**
     * @OA\Get(
     *      path = "/campaigns",
     *      summary = "",
     *      description = "",
     *      operationId = "list",
     *      deprecated = false,
     *     @OA\Response(
     *         response = 200,
     *         description = "successful operation",
     *         @OA\JsonContent(
    *                 allOf = {
    *                     @OA\Schema(ref = "#/components/schemas/ResponseList")
    *                 },
    *                 @OA\Property(
    *                     property = "data",
    *                     ref = "#/components/schemas/ListData",
    *                     @OA\Property(
    *                         property = "items",
    *                         @OA\Items(
    *                             ref = "#/components/schemas/Campaign"
    *                         )
    *                     )
    *                 )
     *         )
     *     )
     * )
     * 
     */

I override the "items" property of the "data" property (ListData object) of my ResponseList object

1. file: Annotations/Schema.php 
moved "public $allOf;" aftter the "public $ref;"

2. file: Annotations/AbstractAnnotation.php
changed the last part of the jsonSerialize function to (quick and dirty fix):

        // $ref
        if (isset($data->ref)) {
            $data2 = new stdClass();
            $dollarRef = '$ref';
            $data2->$dollarRef = $data->ref;
            unset($data->ref);
            
            $props = get_object_vars($data);
            foreach ($props as $k => $v) {
                $data2->$k = $v;
            }
            $data = $data2;
        }

bfanger:
A [reference](https://cswr.github.io/JsonSchema/spec/definitions_references/#reference-specification) should not contain any other properties, just `{ $ref="#/uri" }`


The produced json reflects the incorrect annotations. (swagger-php could do better in detecting and reporting these)

kamil-kielczewski:
allOff not working :(
https://json-schema.org/understanding-json-schema/structuring.html#extending

PR #545, opened 2018-07-03
Update Schema.php

nms666:
changes from #543


bfanger:
`allOf` is at the right place (near the `anyOf` & `oneOf` properties).

PR #546, opened 2018-07-03
Update AbstractAnnotation.php

nms666:
changes from #543


bfanger:
When using a ref, it should not have other properties

Issue #474, opened 2017-12-15
Line break in description support 


soulkoden:
How do write line break in description? 

I have an enum field and I can't write line break for one.

Example 
```
@SWG\Property(
  type="enum",
  enum={1, 2}, 
  description="* `1` - active >>>???<<< * `2` - inactive",
)
```

I tried `\n` and it escaping to `\\n`. 
I tried it:
```
description="* `1` - active
* `2` - inactive"
``` 

and it view as: 
<ul>
<li>`1` - active<ul><li>`2` - inactive</li></ul></li></ul>

I tried it:
```
description="
* `1` - active
* `2` - inactive
"
```

and it view as:
<ul><li><ul>
<li>`1` - active</li><li>`2` - inactive</li></ul></li></ul>

rlanvin:
I'm having the same issue. There is a workaround here: https://github.com/zircote/swagger-php/issues/326 which consists of removing the `*` at the start of line. However, that creates an issue with many text editors that consider it as an invalid DocBlock syntax.

Even Github will display the lines in red, example:

```php
/**
 * description="
- foo
- bar
 * "
 */
```

So if there is a real solution I'm interested as well.

bfanger:
It annoying and it affects swagger-php, but it's an issue in [Doctrine Annotations](https://github.com/doctrine/annotations) library that swagger-php uses.

Please report the issue there.

If you've found a solution or workaround, feel free let us know.

PR #559, opened 2018-07-31
ugly fix #326 #474

### LINKED ISSUE #326: Multi-line annotation issues
I have some annotations that have really long descriptions:

```php
/*
 * @SWG\Definition(
 *   definition="ImageVariants",
 *   description="A key-value list of image variations. The 'original' key should always be present, but other keys ('large', 'small', etc.) may be present as well, depending on the context and object requested.",
 *  ... etc.
 */
```

My IDE (PHPStorm) tends to try and keep line lengths shorter -- or I do it manually to make things more readable -- so I end up with:

```php
/*
 * @SWG\Definition(
 *   definition="ImageVariants",
 *   description="A key-value list of image variations. 
 *     The 'original' key should always be present, but other keys
 *     ('large', 'small', etc.) may be present as well, 
 *     depending on the context and object requested.",
 *  ... etc.
 */
```

swagger-php can parse this just fine, but the resulting json includes the asterisks from the docblock:

```json
"ImageVariants": {
    "description": "A key-value list of image variations.\n     *     The 'original' key should always be present, but other keys\n     *     ('large', 'small', etc.) may be present as well,\n     *     depending on the context and object requested.",
}
```

Is there anyway around this?  Some way to make long values readable in the docblocks, but still to generate the "correct" JSON?

savage4pro:
handle line breaks as expected (preg_replace)
support cyrillic as is (JSON_UNESCAPED_UNICODE)

example
```
 * @OA\Info(
 *     version="1.0",
 *     title="Sample API",
 *     description="Длинное\n\nпредлинное
 *     `1` first
 *     `2` second
 *     ```shell\nX-Request-ID: g9h3LkdJf8\n```
 *     Описание api
 *     для блока ```@OA\Info```",
 *     @OA\Contact(
 *         name="company name",
 *         url="https://domain.tld"
 *     ),
 * )
```
generates
```
    "info": {
        "title": "Sample API",
        "description": "Длинное\n\nпредлинное\n\n`1` first\n\n`2` second\n\n```shell\nX-Request-ID: g9h3LkdJf8\n```\n\nОписание api\n\nдля блока ```@OA\\Info```",
        "contact": {
            "name": "company name",
            "url": "https://domain.tld"
        },
        "version": "1.0"
    },
```
and in e.g. [ReDoc](https://github.com/Rebilly/ReDoc) it looks something like this
ReDoc Sample Info

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -222,7 +222,7 @@ abstract class AbstractAnnotation implements JsonSerializable
 
     public function __toString()
     {
-        return json_encode($this, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES);
+        return preg_replace(["/\\\\\\\\[nr]/", "/(\\\\[nr])+\s+\*\s+/"], ['\n','\n\n'], json_encode($this, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE));
     }
 
     public function __debugInfo()


bfanger:
> support cyrillic as is (JSON_UNESCAPED_UNICODE)

I agree, i've merged this part of the PR as  a08d4b7  

> handle line breaks as expected (preg_replace)

I'm still hoping doctrine-annotations would implement beter multiline support.
For now you need to remove the leading ` *  ` which some editors (and most programmers) don't like.

You could of course always run the preg_replace outside of swagger-php

```php
require("vendor/autoload.php");
$openapi = \OpenApi\scan('/path/to/project');
header('Content-Type: application/json');
echo preg_replace(["/\\\\\\\\[nr]/", "/(\\\\[nr])+\s+\*\s+/"], ['\n','\n\n'], $openapi);
```

PR #561, opened 2018-08-02
Allow the use of the discriminator inside property and items object

penja:
According to doc we can use `oneOf` inside properties and items. For more clearness, oneOf is also used with a discriminator.
```
type: array
items:
  oneOf:
    - $ref: '#/components/schemas/Cat'
    - $ref: '#/components/schemas/Dog'
```
This PR allows you to use the `Discriminator` class inside the `Property` and `Items` classes.


bfanger:
Thanks! merged as  5a47337

PR #562, opened 2018-08-03
Don't generate description for property with ref 

penja:
According to this https://github.com/OAI/OpenAPI-Specification/issues/556 issue we should't generate description for properties with ref.


bfanger:
Merged as 47cbb3b, I've changed it so instead of the oddly specific description, it doesn't auto-add **any** properties when the ref is set.

penja:
@bfanger Thanks

PR #563, opened 2018-08-07
Set example as null if property is nullable but no examples have been defined

vitxd:
Hello there

this might not be the best solution but it suits my needs. I'm very happy to change the way it does this with the lead of someone more experienced in this project.

My first idea was to create a null object to use or wrap null in strings and detect that but in the end this is working fine.

Let me know your thoughts


bfanger:
For cases when the value can be null, the value should be set to the UNDEFINED constant in the class. Seems we missed the example property.

bfanger:
All properties are now set to the special `UNDEFINED` value, this allows the annotation to export `null` values. `@OA\Schema(example=null)` now works.

Issue #556, opened 2018-07-26
The InheritProperties does not work as expected


penja:
@bfanger The [InheritProperties processor](https://github.com/zircote/swagger-php/blob/master/src/Processors/InheritProperties.php) that was added by proposal https://github.com/zircote/swagger-php/issues/348 does not work correctly with `allOf` schema property. 
Example.
```
namespace OpenApiFixtures;

/**
 * @OA\Schema()
 */
class BaseModel {

    /**
     * @OA\Property();
     * @var string
     */
    public $baseProperty;
}
```
And
```
/**
 *  @OA\Schema(
 *   schema="ExtendedModel",
 *   allOf={
 *      @OA\Schema(ref="#/components/schemas/BaseModel"),
 *   }
 * )
 */
class ExtendedModel extends BaseModel
{

    /**
     * @OA\Property();
     * @var string
     */
    public $extendedProperty;
}
```
Generate 
```
{
	"schema": "BaseModel",
	"properties": {
		"baseProperty": {
			"type": "string"
		}
	}
}
```
And
```
{
	"schema": "ExtendedModel",
	"properties": {
		"baseProperty": {
			"type": "string"
		}
	},
	"allOf": [{
		"properties": {
			"extendedProperty": {
				"type": "string"
			}
		},
		"$ref": "#\/components\/schemas\/BaseModel"
	}]
}
```
As you see, we have an extra property `baseProperty` in `ExtendedModel`.

We must not inherit the properties of the base class if these properties are inherited by a openapi. That fields should be in existing list in `InheritProperties` processor.

Can you tell how get the annotation object `BaseModel` by reference `#/components/schemas/BaseModel`? And I will prepare a PR that will solve this problem.

bfanger:
I suggest to following  behavior of the new InheritProperties processor:

- If the parent class doesn't have an annotation then inject/copy the properties into the  child (current behavior).

- If the parent class has an annotation and the allOf in the child is empty, create the allOf with a $ref to the parent.
  
- If the parent class has an annotation and the allOf in the child is set, do nothing and allow the author the flexibility to set it anything including an `[]`, perhaps hide the allOf in the output if it's empty.

> Can you tell how get the annotation object BaseModel by reference.

This is tricky but possible, you can get the annotation and look at its `_context`, this should allow us to generate the ref when the annotation has been placed in the tree. 
[...]

PR #564, opened 2018-08-08
Fix for InheritProperties for #556

penja:
PR for https://github.com/zircote/swagger-php/issues/556

- If the parent class has an annotation and the allOf in the child is empty, create the allOf with a $ref to the parent.
- If the parent class has an annotation and the allOf in the child is set, do nothing
- If child class has two schemas only in first schema will inherit property
- Don't merge properties into nested schemas


bfanger:
Thanks!

PR #575, opened 2018-08-24
A class using a trait will include it's annotated properties automatically

dafrost22:
Adding to #524 with what was in my older version pull request #522 in order to automatically merge the properties of a class with the properties of a trait it is using


dafrost22:
@bfanger I'm retrofitted what I had working from before. Maybe @gjergjsheldija should weigh in here too... In InheritProperties.php I'm trying to combine the properties of classes and traits to become one model, but I needed my own function in Analysis.php (getUsedTraits) to get them, since the one from the merged PR (getTraitsOfClass) isn't returning traits with the classes. Maybe there is a special way to set up the annotations on classes and traits to get them to combine properly I'm missing? With the above changes it's automatic, but either way is fine with me as long as I can get it to work. Thanks all!

src/Analysis.php
```diff
@@ -219,6 +219,23 @@ public function getTraitsOfClass($class)
         return $definitions;
     }
 
+    public function getUsedTraits($class)
+    {
+        if ($class == '\\' || !class_exists($class)) {
+            return [];
+        }
+        $usedTraits = class_uses($class);
```
bfanger: This only works when the class/trait can be autoloaded.
When swagger-php cli is installed globally it doesn't execute any php code in the target project. Therefor we can reliably use reflection or class_exits etc.

dafrost22:
@bfanger Understood about the class_uses, So instead I looked for any issues in the master branch to see why I was not able to see traits using the in-place functions, and posted a fix in a separate pull request. Now this pull request really only addresses the automatic merging of properties of traits and I wanted to see if it's something you would consider. 

Currently it looks like a trait must have a Schema annotation, which would make it its own model inside of the OpenAPI JSON output, and if another class references it, it does an "allOf" reference if the Model of the Class using the trait. If you do not include the Schema annotation at the top, an error happens saying that the Property annotations within the trait must be inside another annotation (like Schema, etc, forcing a Schema annotation to be put in place). My question now is what if I do not want to include a trait as a Model in the JSON output, and would rather it be hidden, so that the properties just appear to be a part of the Model of my class rather than a reference to a whole other object/model in the JSON. 

The changes in this pull request are now showing only how to automatically merge the properties together. If this is an option you'd like to explore, I believe the thing to do would be to only combine the properties if the trait does NOT have a Schema annotation of its own but does have Property annotations, and those properties become part of the Class' Model that is using the trait. Otherwise, if there is a Schema annotation on the trait, it behaves as it does already, and creates its own Model in the JSON output, which can then be referenced by other Models.

Hope I'm making sense, I wanted to get your opinion first before I went further. Thanks,

dafrost22:
I spoke too soon, I may have had some conflicting changes making the behavior I described when testing. I have discovered that with @gjergjsheldija 's commit, it does in fact combine the properties like I would have intended, but a fix needed to be made for blank OA\Property() annotations. Closing this, please look at PR #576

PR #576, opened 2018-08-27
Trait reference fix

dafrost22:
The 'trait' array of class definitions would not populate if referencing a trait by a different namespace.

This would attach to 'traits':
`use Hello;`
But this would not
`use Subfolder\Hello`


dafrost22:
Blank annotations like:
`@OA\Property()`
Now are being accounted for when merging properties with a class

gjergjsheldija:
hi @dafrost22 
if i understood correctly the bug was that empty properties where not getting merged ?

regards

gjergjsheldija:
@dafrost22 thank you for the kind words
regarding the PR, i must have probably missed the empty Property() because it wasn't in the test case. thank you for that :)
regarding the traits thing, it was intentional, but i can see the general case for it.
thanks

lopezdonaque:
Hi, 
Is this PR merged into current 3.0 release?
I'm trying to use traits and I get the next error if the Schema annotation is not defined in the trait.

`Unexpected @OA\Property(), expected to be inside @OA\AdditionalProperties, @OA\Schema, @OA\JsonContent, @OA\XmlContent, @OA\Property, @OA\Items in app/Hexagonal/Domain/Api/Base/Resources/Properties/PropertySpecificationsTraitResource.php on line 125`

If I use the Schema annotation in the trait, no error is triggered but the result contains two schemas.
The first schema (the class which uses the trait) only contains the first property of the trait.
The second schema (the trait) contains all the properties excepts the first one (included in the first schema).

Thanks

lopezdonaque:
I have used @dafrost22 branch in composer since it will be merged.

`    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/dafrost22/swagger-php.git"
        }
    ]`

`"zircote/swagger-php": "dev-trait-reference-fix"`

zim32:
Still not merged?

laszlof:
Can we get this merged? I also ran into this issue.

bfanger:
The current traits support I merged buggy, but this PR doesn't add proper namespace support either.

PR #582, opened 2018-09-16
Add support to import traits from nested classes

gjergjsheldija:
With the old trait implementation there was a bug with traits that where used inside other classes, the new traits test should clarify it a bit better and give and example of the issue. This pull should fix it


itsjavi:
any idea when is this going to be released? I have problems when using traits with hierarchies

gjergjsheldija:
@itsjavi sorry, no idea

bfanger:
I've removed the unused files and merged the PR as f10ab7f

PR #584, opened 2018-09-19
Add var type detection when null is also set

DSB:
- add removing of "null" at type detection
- add test cases

### DIFF
--- a/src/Processors/AugmentProperties.php
+++ b/src/Processors/AugmentProperties.php
@@ -65,7 +65,7 @@ class AugmentProperties
             if (preg_match('/@var\s+(?<type>[^\s]+)([ \t])?(?<description>.+)?$/im', $context->comment, $varMatches)) {
                 if ($property->type === UNDEFINED) {
                     preg_match('/^([^\[]+)(.*$)/', trim($varMatches['type']), $typeMatches);
-                    $type = $this->exractType($typeMatches[1]);
+                    $type = $this->extractType($typeMatches[1]);
                     if (array_key_exists(strtolower($type), static::$types) === false) {
                         $key = strtolower($context->fullyQualifiedName($type));
                         if ($property->ref === UNDEFINED && $typeMatches[2] === '' && array_key_exists($key, $refs)) {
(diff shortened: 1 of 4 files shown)


src/Processors/AugmentProperties.php
```diff
@@ -113,4 +112,26 @@ public function __invoke(Analysis $analysis)
             }
         }
     }
+
+    /**
+     * @param string $type
```
Tobion: the phpdoc does not add anything usefull and does not have the correct var name. imo you can remove the phpdoc

src/Processors/AugmentProperties.php
```diff
@@ -113,4 +112,26 @@ public function __invoke(Analysis $analysis)
...
+     *
+     * @return string
+     */
+    protected function exractType(string $typeString)
```
Tobion: why $typeString when the variable already has a type declarations. just name it $type. you can also add a return type

src/Processors/AugmentProperties.php
```diff
@@ -113,4 +112,26 @@ public function __invoke(Analysis $analysis)
...
+
+        $types = explode('|', $typeString);
+        foreach ($types as $type) {
+            $type = trim($type);
+            if ($type !== 'null') {
+                return $type;
+            }
+        }
+
+        return 'UNKNOWN';
```
Tobion: why UNKNOWN? does it have special meaning?

src/Processors/AugmentProperties.php
DSB: Because in foreach I also use types as $type and I don't want to confuse the reader with a var named the same but with different content. With typeStrign I mean "The string which contains the type in it but also holds more info". I will think about a better naming.

src/Processors/AugmentProperties.php
DSB: Ok, I will update the pr.

src/Processors/AugmentProperties.php
DSB: It is not in the mapping of known types and triggers the right behavior. I will try to change it to the UNDEFINED constant.

src/Processors/AugmentProperties.php
DSB: I simply reused the given naming. ;)

DSB:
I updated the pull request according to your comments. Will someone review it?

bfanger:
Thanks! 
Tweaked and merged as 68674610a9951494d86da9f2c6d0ef4aa5987ea0

PR #585, opened 2018-09-21
fix for allOf in InheritProperties

penja:
We should add allOf for child schema only if parent schema has defined schema name. This behavior has been broken when we added undefined as default value into all annotation properties.


penja:
@bfanger Сan you look into and merge this small fix ?

bfanger:
Merged as fd9591d17b937cd0c8818abd25f90b87024694b3

Thanks!

PR #603, opened 2018-10-21
integration with docker automated build

axklim:


bfanger:
Thanks, but i'm not merging this PR because that would make me the maintainer of the docker support.

PR #610, opened 2018-11-02
Add http response codes range definitions support with tests

ValentinNikolaev:
As you know swagger-php throw notice if http code range definitions was found (or something not "default" or a HTTP Status Code). This is incorrect behavior. We should have support of the range defenitions (check https://swagger.io/docs/specification/describing-responses/, part **HTTP Status Codes**)

Yeah, in fact still have valid yml/json but with notice in log. I.e.
```
Invalid value "4XX" for @OA\Response()->response, expecting "default" or a HTTP Status Code in modules/Http/Controllers/DefaultController.php on line 33
```


ValentinNikolaev:
@bfanger

PR #618, opened 2018-11-28
Configurable File Patterns added

ktamas77:
Configurable File Patterns added for OpenAPI (CLI included).
This enables to properly scan projects which are using alternate extensions for PHP files.
The pattern parameter is using standard Symfony/Finder patterns.


bfanger:
Thanks! I've merged it as 5a697be

PR #622, opened 2018-12-04
Update README.md

dcrouch79:
Fixed broken link to doctrine annotations website

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # swagger-php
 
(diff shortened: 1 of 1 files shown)


bfanger:
swagger-php tries to be compatible with latest version of doctrine.

So linking to latest is fine, the warning at the top of the documentation is enough.
Which version in installed also depends on the requirements of other packages.

Thanks for fixing the link.

PR #623, opened 2018-12-06
Optimized scan & Configurable file patterns

ktamas77:
- Dynamic swagger generation is now pre-scans files for OpenApi definitions before it's passing it to the Tokenizer in order to speed up dynamic API generation for large repositories
- Configurable Patterns added


src/functions.php
```diff
@@ -17,12 +17,24 @@
...
 
+    /**
+     * Determines whether the given string has openapi definitions or not
+     *
+     * @param string $filecontent File content to analyze
+     * @return boolean
+     */
+    function hasOpenApiDefinitions($filecontent)
+    {
+        return strpos($filecontent, '@OA\\') !== false;
```
penja: This will not work for the custom annotations under other namespaces

ktamas77: true - i'll close this PR

penja: You can try get info from `Analyser::defaultImports` and `Analyser:whitelist` properties. But we should investigate this and coverage all cases with tests.

bfanger:
Some features like inheritance rely on parsing intermediate files which might not have annotations. 
But If your project doen't, using **grep** with **xargs** you can apply the same optimisation:

```sh
grep --null -r -l --include "*.php"  "@OA" $my-project-directory | xargs -0 openapi
```
I consider swagger-php fast enough, and therefor closing this PR, but thanks for the ideas.

PR #631, opened 2018-12-14
Improve XmlContent

hrab:
The XmlContent annotation is ignored in RequestBody. Also, defining a single example is not supported.

These both work for JsonContent annotation. This pull request is to bring XmlContent up to par with JsonContent.


bfanger:
Thanks! merged as ad70f2ae659ac4bb50db96091a68da06c504b5a6

PR #635, opened 2018-12-21
update openapi to solve the iteration problem

gbabyX:
the for loop increase $i,if do not delete this '$i++',it would ignore the next arg


bfanger:
This patch is not bug free, what issue were you trying to solve?

PR #643, opened 2019-01-07
Add var type detection when null is also set 

thePanz:
Backport of 6867461

### DIFF
--- a/src/Processors/AugmentProperties.php
+++ b/src/Processors/AugmentProperties.php
@@ -8,6 +8,7 @@ namespace Swagger\Processors;
 
 use Swagger\Annotations\Definition;
 use Swagger\Annotations\Items;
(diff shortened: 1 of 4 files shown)


Tobion:
@bfanger could you merge this please and potentially release a new 2.x version?

Tobion:
What I've found is that `@var string|null` and `@var MyClass|null` works now as the tests also prove. But `@var MyClass[]|null` does not work while `@var MyClass[]` itself works. So there seems to be a missing case for the `|null` handling.

thePanz:
Thanks @bfanger . Could you tag a new release on the 2.x branch?

thePanz:
@bfanger can you tag a new release with the changes from this PR? thank you :)

PR #644, opened 2019-01-08
Unknown Parameter error message update:

dcarbone:
- Adding small func "debugVarOutput" to help with converting var types to error-message focused value
- Adding some additional context output to the "unknown parameter" error in AbstractAnnotation
- Adding test case


DerManoMann:
I like the idea of better error messages, but OTOH I am not a fan of global functions.
Maybe this could go into the `Util` class (which should be `Utils`, IMHO)?

bfanger:
This change would cause very large error messages. My philosophy is that the messages should be succinct: Be as short as possible but not shorter. 

We could look into upgrading the identity() methods to display a little more information to identify which annotation was the problem.

dcarbone:
@bfanger how exactly does this equate to large error messages?  i'd like to see proof of this, because as it stands the error message is useless.

bfanger:
Do you have a specific error scenario in mind?

bfanger:
Example:

`Unexpected field "titel" for @OA\Property(property="config"), expecting "property", "ref", "schema", "title", "description", "maxProperties", "minProperties", "required", "properties", "type", "format", "items", "collectionFormat", "default", "maximum", "exclusiveMaximum", "minimum", "exclusiveMinimum", "maxLength", "minLength", "pattern", "maxItems", "minItems", "uniqueItems", "enum", "multipleOf", "discriminator", "readOnly", "writeOnly", "xml", "externalDocs", "example", "nullable", "deprecated", "allOf", "anyOf", "oneOf", "not", "additionalProperties", "additionalItems", "contains", "patternProperties", "dependencies", "propertyNames", "const", "x" in /Users/bob/test.php on line 6`

Which [additional work](https://www.php.net/manual/en/function.levenshtein.php) this message could become:


`Unexpected field "titel" for @OA\Property(property="config"), did you mean "title"? in /Users/bob/test.php on line 6`

PR #648, opened 2019-01-18
Fix property value in Examples is only string

dizairo:
Property value is limited to a string, although an example can also be represented by a boolean, integer, array or object.


bfanger:
Thanks! Merged as ff39699

PR #659, opened 2019-03-12
Update SecurityScheme.php

herokiller:
currently scheme is required field, not securityScheme


bfanger:
Closing this PR because:
- It broke the unittests
- Didn't describe the problem it tries to solve.

Issue #670, opened 2019-04-12
How to handle nullable property references?


BenMorel:
First of all sorry if I missed something obvious, but I've been struggling to find any information about how to handle this basic use case.

I have a Schema class that's not defined as `nullable`:

```php
/**
 * @OA\Schema()
 */
class Deal
{
}
```

When I want to use it in another Schema class and make it nullable, I write:

```php
/**
 * @OA\Schema()
 */
class Reservation
{
    /**
     * @OA\Property()
     *
     * @var Deal|null
     */
    public $deal;
}
```

However, `|null` is not taken into account like it is for scalar types, and this generates a `$ref` to the non-nullable Schema, like if I'd written `@var Deal`:

```yaml
components:
  schemas:
    Reservation:
      properties:
        deal:
          $ref: '#/components/schemas/Deal'
```

**Is this expected behaviour?** I'd expect this library to generate something like:

```yaml
        deal:
          nullable: true
          allOf:
            -
              $ref: '#/components/schemas/Deal'
```

Which is [what the spec seems to suggest](https://github.com/OAI/OpenAPI-Specification/issues/1368).

In the meantime, I tried to manually configure the `@Property`, with no luck so far:

```php
    /**
     * @OA\Property(allOf={@OA\Schema(ref="#/components/schemas/Deal")}, nullable=true)
     *
     * @var Deal|null
     */
    public $deal;
```

This does not work as it still generates the `$ref`:

```yaml
        deal:
          nullable: true
          allOf:
            -
              $ref: '#/components/schemas/Deal'
          $ref: '#/components/schemas/Deal'
```

**What did I miss?** Thanks in advance!

BenMorel:
After digging through the code, it looks to me like this case is just not supported at the moment.
So I went ahead and implemented it in #671.

Feedback welcome, I desperately need this feature! 👍

bfanger:
Closing this as "duplicate" of #671

PR #671, opened 2019-04-12
Implement nullable refs

BenMorel:
This is a follow-up to #670.

The idea is to transform a nullable property:

```php
/**
 * @OA\Schema()
 */
class Reservation
{
    /**
     * @OA\Property()
     *
     * @var Deal|null
     */
    public $deal;
}
```

Into this:

```yaml
components:
  schemas:
    Reservation:
      properties:
        deal:
          nullable: true
          allOf:
            -
              $ref: '#/components/schemas/Deal'
```

The implementation is simple:

- if the property types contain 'null': set `nullable` and `allOf` with `ref`
- else: set `ref` as usual

✅ No BC breaks
✅ Covered by tests

### DIFF
--- a/src/Processors/AugmentProperties.php
+++ b/src/Processors/AugmentProperties.php
@@ -66,11 +66,22 @@ class AugmentProperties
             if (preg_match('/@var\s+(?<type>[^\s]+)([ \t])?(?<description>.+)?$/im', $comment, $varMatches)) {
                 if ($property->type === UNDEFINED) {
                     preg_match('/^([^\[]+)(.*$)/', trim($varMatches['type']), $typeMatches);
+                    $isNullable = $this->isNullable($typeMatches[1]);
                     $type = $this->stripNull($typeMatches[1]);
                     if (array_key_exists(strtolower($type), static::$types) === false) {
                         $key = strtolower($context->fullyQualifiedName($type));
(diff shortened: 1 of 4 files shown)


BenMorel:
Any feedback here?

BenMorel:
Bump.

bfanger:
I noticed the `allOf` (which is correct btw) but I'd like to see `oneOf` and general support for the `|` operator, not just for `null`, for example:

```
/**
 * @var Deal|Scam
 */
```
Would generate:

```yaml
       deal:
          oneOf:
            -
              $ref: '#/components/schemas/Deal'
            -
              $ref: '#/components/schemas/Scam'
```

Do you agree? and would you like to implement that feature?

BenMorel:
I agree this would be nice, but I think that it is a separate issue that can be handled after the current PR is merged! I can work on it if you wish.

bfanger:
Merged as 615941e42d31dda89add9508c9321daac938846a 
(I've changed `allOf` to `oneOf`, with only one entry it has the same meaning, but makes the aforementioned feature easier to implement)

Looking forward to your PR

BenMorel:
Cool, thanks! I'll try to find some time this week to work on the generic union types.

PR #676, opened 2019-04-30
Pet ID copypasta fix

SeanJA:
Pet ID was copied to multiple properties as the title, I am assuming it was by accident

### DIFF
--- a/Examples/petstore-3.0/models/Order.php
+++ b/Examples/petstore-3.0/models/Order.php
@@ -61,7 +61,6 @@ class Order
      *     format="datetime",
      *     description="Shipping date",
      *     title="Shipping date",
-     *     title="Pet ID",
      *     type="string"
      * )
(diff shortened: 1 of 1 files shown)


SeanJA:
Wait... was that on purpose to make sure that it would take the last one on the list?

bfanger:
The Petstore example was to test if swagger-php 2.x was feature-complete and could generate the `swagger.json` from https://petstore.swagger.io/

(Including any spelling and other mistakes)

Sadly the petstore isn't updated to te latest spec (v3 / OpenApi) so right now its in a weird state, but still useful as an example of how to document an api.

Fixed unittest and merged as 5677ea2c9b40035bd32f65197246fc3f28c7cbc4

SeanJA:
Cool, I didn't realize that until the tests failed 👍

PR #693, opened 2019-07-06
Failure test for Analysis::merged implementation

### LINKED ISSUE #678: Potential bug in Analysis::unmerged()
https://github.com/zircote/swagger-php/blob/master/src/Analysis.php#L308-L310

These lines looks strange. I'm trying to use `unmerged` but get all unmerged annotations back into resulting analysis. It looks like this code utilizes the same `OpenAPI` by reference instance and sets `_umerged` back into it, so it does no effect

scaytrase:
Exposes and fixes #678

### DIFF
--- a/src/Analysis.php
+++ b/src/Analysis.php
@@ -333,7 +333,7 @@ class Analysis
         $result->merged = $this->merged();
         $result->unmerged = new Analysis();
         foreach ($this->annotations as $annotation) {
-            if ($result->merged->annotations->contains($annotation) === false) {
+            if ($annotation instanceof OpenApi === false && $result->merged->annotations->contains($annotation) === false) {
                 $result->unmerged->annotations->attach($annotation, $this->annotations[$annotation]);
             }
         }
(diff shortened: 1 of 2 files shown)


scaytrase:
Failure is intended and demostrates the bug

scaytrase:
I've also proposed simple fix for this bug in separate commit.

Fixes #678

bfanger:
Excellent! I'll look into it when i can schedule some time.

src/Analysis.php
```diff
@@ -334,7 +333,7 @@ public function split()
         $result->merged = $this->merged();
         $result->unmerged = new Analysis();
         foreach ($this->annotations as $annotation) {
-            if ($result->merged->annotations->contains($annotation) === false) {
+            if ($annotation instanceof OpenApi === false && $result->merged->annotations->contains($annotation) === false) {
```
scaytrase: This is because we cannot check different instances of `OpenAPI` now to be in `annotations` storage directly. Not sure if it is correct fix, but it helped

DerManoMann:
Should this be closed since the issue it tries to fix is also closed? Also, the PR is badly out of date...

Not sure I understand the problem at all, though; the issue talks about trying to use `unmerged()` but the test calls `merge()`...

bfanger:
I'm closing this as i've not seen any bugreport related to this and i'm not experiencing any issues, but we'll keep this PR in mind when we're updating this part of the code.

PR #605, opened 2018-10-29
Fix for deserialization of allOf properties

penja:
Deserializer work incorrect when we trying to deserialize schema with `allOf` property.
For example allOf properties will be deserialized like `stdClass` with the `$ref` property but according to swagger `allOf,oneOf,anyOf` is array of schemes [link](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/).

Json  string example
```
{
	"openapi": "3.0.0",
	"info": {
		"title": "Pet store",
		"version": "1.0"
	},
	"components": {
		"schemas": {
			"Dog": {
				"allOf": [{
					"$ref": "#/components/schemas/SomeSchema"
				}]
			},
			"Cat": {
				"allOf": [{
					"$ref": "#/components/schemas/SomeSchema"
				}]
			}
		}
	}
}
```
The result of deserealization
```
object(OpenApi\Annotations\OpenApi)#2844 (5) {
  ["openapi"]=>
  string(5) "3.0.0"
  ["info"]=>
  object(OpenApi\Annotations\Info)#2848 (4) {
    ["title"]=>
    string(9) "Pet store"
    ["version"]=>
    string(3) "1.0"
    ["_context"]=>
    object(OpenApi\Context)#2846 (1) {
      ["-"]=>
      string(107) "\OpenApi\Serializer->doDeserialize() in .../dev/swagger-php/src/Serializer.php on line 120"
    }
    ["_unmerged"]=>
    array(0) {
    }
  }
  ["components"]=>
  object(OpenApi\Annotations\Components)#2847 (3) {
    ["schemas"]=>
    array(2) {
      ["Dog"]=>
      object(OpenApi\Annotations\Schema)#2798 (4) {
        ["schema"]=>
        string(3) "Dog"
        ["allOf"]=>
        array(1) {
          [0]=>
          object(stdClass)#2852 (1) {
            ["$ref"]=>
            string(31) "#/components/schemas/SomeSchema"
          }
        }
        .....
```
My proposal to mark the properties that hasn't simple key like discriminator or items as simple array of objects. This PR will fix this problem also this fix is covered by the unit test.

PR #699, opened 2019-07-17
fix for deserialization of allOf property (2.x)

IMM0rtalis:
partly backport of #605 

Partly because we just have `allOf` available as property right now, no `oneOf` or anyOf.
Also I did not do other refactoring which was done in #605 to not change to much in 2.x

Simple test for definitions is added


IMM0rtalis:
@bfanger can you look at this PR when you can spare some time? The `allOf` not properly deserializing annoys me for a long time and this could fix it finally

bfanger:
I'm no longer supporting 2.x, i barely have spare time for 3.x

Issue #705, opened 2019-08-05
Nullable property


xLs51:
For the following property : 

```
/**
 * The ID.
 * @OA\Property()
 * @var int|null
 */
private $id;
```

I would expect : 

```
properties:
  id:
    description: 'The ID.'
    type: integer
    nullable: true
  type: object
```

But I get : 

```
properties:
  id:
    description: 'The ID.'
    type: integer
  type: object
```

I can of course fix it by adding `@OA\Property(nullable=true)`, but couldn't this be done automatically?

bfanger:
Yes, that is the wanted behaviour, similar to https://github.com/zircote/swagger-php/pull/671 

Could you create a PR adding this feature?

PR #711, opened 2019-08-26
Implement nullable properties

xLs51:
Fix: https://github.com/zircote/swagger-php/issues/705


range-of-motion:
Dope, need this :+1:

range-of-motion:
@zircote this would be really useful to have. When can we expect this in a release?

bfanger:
Merged as f1e9122f35e0c2ed343c2fd0dc4c322ed7a49f35

PR #714, opened 2019-08-29
[interface-parsing] parse interfaces, just like classes and traits

### LINKED ISSUE #713: Parsing interfaces (just like classes and traits)
I would like to be able to parse interfaces, just like you can classes and traits.

This helps with making extendable code, writing an interface with the PHPDocs on it, where each consumer of the interface can leverage the same phpdoc, but write their own implementation.

Doqnach:
Fixes #713 

Might be missing more tests?

### DIFF
--- a/src/Analysis.php
+++ b/src/Analysis.php
@@ -8,6 +8,7 @@ namespace OpenApi;
 
 use Closure;
 use Exception;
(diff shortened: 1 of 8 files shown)


Doqnach:
@bfanger any chances you could look at this PR? Would be nice to be able to switch back to the main repo, from my own fork.

PR #722, opened 2019-10-01
PHP 7.4 Support

sunspikes:
Adds PHP 7.4 Support

PR #736, opened 2019-11-30
Fix PHP 7.4 notice for invalid array access

andreasschroth:
next() returns false if there are no more elements left and thus an invalid array access occurs (on the returned false value).
The exact notice is: "Trying to access array offset on value of type bool"

For more information see this RFC: https://wiki.php.net/rfc/notice-for-non-valid-array-container


andreasschroth:
Closing in favor of https://github.com/zircote/swagger-php/pull/722 .

PR #741, opened 2019-12-10
Add PHP7.4 compatibility for those who still use OpenApi2

eugene-chernogradski:


bfanger:
Thanks! merged and released as 2.0.15

PR #743, opened 2019-12-15
Fix typed properties in php 7.4

### LINKED ISSUE #742: PHP 7.4 Class typed properties breaks Analysis
Hi,

The new PHP 7.4 typed property feature is a godsend for this library and would really help reduce the use of comment Annotations in Models.

`type` can now be directly deduced from `PHP` tokens as well as the `nullable` property (c.f. #705 / #711), which brings a more robust way of defining Models.

Example:
PHP < 7.4
```php
/**
 * @OA\Schema
 */
class ExampleModel
{
    /**
     * @var int|null
     * @OA\Property
     */
    private $id;

    /**
     * @var string
     * @OA\Property
     */
    private $name;

    /**
     * @var \DateTimeInterface|null
     * @OA\Property(type="datetime")
     */
    private $createdAt;
}
```

PHP >= 7.4
```php
/**
 * @OA\Schema
 */
class ExampleModel
{
    /** @OA\Property */
    private ?int $id;

    /** @OA\Property */
    private string $name;

    /** @OA\Property(type="datetime") */
    private ?\DateTimeInterface $createdAt;
}
```

I tried them out and they make the `StaticAnalyser` fail; it fails to get property names while parsing the tokens as it doesn't recognise the new format.

jpuck:
Fixes #742

Additionally, this allows the parser to now read [php 7.4 native property types][typed-properties] for augmenting the schema, so no need to declare `@var Type` anymore 🎉 

[typed-properties]:https://www.php.net/manual/en/migration74.new-features.php#migration74.new-features.core.typed-properties

### DIFF
--- a/src/StaticAnalyser.php
+++ b/src/StaticAnalyser.php
@@ -35,7 +35,7 @@ class StaticAnalyser
                 $GLOBALS['openapi_opcache_warning'] = true;
                 $status = opcache_get_status();
                 $config = opcache_get_configuration();
-                if ($status['opcache_enabled'] && $config['directives']['opcache.save_comments'] == false) {
+                if (is_array($status) && $status['opcache_enabled'] && $config['directives']['opcache.save_comments'] == false) {
                     Logger::warning("php.ini \"opcache.save_comments = 0\" interferes with extracting annotations.\n[LINK] http://php.net/manual/en/opcache.configuration.php#ini.opcache.save-comments");
                 }
             }
@@ -326,17 +326,19 @@ class StaticAnalyser
     {
         while (true) {
             $token = next($tokens);
-            if ($token[0] === T_WHITESPACE) {
-                continue;
-            }
-            if ($token[0] === T_COMMENT) {
-                $pos = strpos($token[1], '@OA\\');
-                if ($pos) {
-                    $line = $context->line ? $context->line + $token[2] : $token[2];
-                    $commentContext = new Context(['line' => $line], $context);
-                    Logger::notice('Annotations are only parsed inside `/**` DocBlocks, skipping ' . $commentContext);
+            if (is_array($token)) {
+                if ($token[0] === T_WHITESPACE) {
+                    continue;
                 }
-                continue;
+                if ($token[0] === T_COMMENT) {
+                    $pos = strpos($token[1], '@OA\\');
+                    if ($pos) {
+                        $line           = $context->line ? $context->line + $token[2] : $token[2];
+                        $commentContext = new Context(['line' => $line], $context);
+                        Logger::notice('Annotations are only parsed inside `/**` DocBlocks, skipping ' . $commentContext);
+                    }
+                    continue;
+                }
             }
 
             return $token;
--- a/src/StaticAnalyser.php
+++ b/src/StaticAnalyser.php
@@ -187,6 +187,9 @@ class StaticAnalyser
                 if ($token[0] == T_STATIC) {
                     $token = $this->nextToken($tokens, $parseContext);
                 }
+                if ($token === '?') { // nullable type
+                    $token = $this->nextToken($tokens, $parseContext);
+                }
                 if ($token[0] === T_STRING) { // property type declaration
                     $type = $token[1];
                     $token = $this->nextToken($tokens, $parseContext);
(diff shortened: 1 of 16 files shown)


src/StaticAnalyser.php
```diff
@@ -391,4 +389,25 @@ private function parseUseStatement(&$tokens, &$token, $parseContext)
 
         return $statements;
     }
+
+    private function extractTypeAndNextToken(array &$tokens, Context $parseContext): array
+    {
+        $type = UNDEFINED;
```
ymilin: When using the FQCN notation on non-primitive types like `\DateTimeInterface`, I think the character `\` will be in the tokens array as a character like `?`.

jpuck: oh snap, you're right. moreover, we will have to handle multi-segment namespaces too, e.g. `\App\Models\Foo` because that would tokenize as `T_NS_SEPARATOR` . `T_STRING` . `T_NS_SEPARATOR` . `T_STRING` . `T_NS_SEPARATOR` . `T_STRING`

We should probably abstract something out of `\OpenApi\StaticAnalyser::parseUseStatement` since much of that implementation logic can be reused here.

I will put this PR back into draft now and try to fix that up tonight.

jpuck: all right that should fix it 👍

ymilin:
Wow, awesome work! Thank you so much!

LGTM and can't wait to use it 😄

Bavragor:
Would be great to see this merged. :+1:

rsgrey:
Pls, merge!

Adrians-G:
Pls, merge!!!!

jpuck:
@darkbluesun good catch! It looks like the formatting of oas types from native types/docs happens somewhere else. I thought I could hook in here and take advantage of that. Lemme take a look at this in the next couple days. Also happy to take pull requests into my branch

darkbluesun:
@jpuck here's my best shot at it. Take or leave https://github.com/jpuck/swagger-php/pull/1

jpuck:
@bfanger would you please take a look at this one when you get a chance? would appreciate any feedback. thank you in advance.

SlvrEagle23:
@jpuck Just switched over to using this fix branch for building our API documentation, as we've been using PHP 7.4's typed properties for a while now.

It seems to work with no issues, with the exception of properties with the `array` type, which get incorrectly assigned as properties with the name `''`. I noticed this from multiple arrays triggering a PHP warning saying "multiple properties with the name ''" when generating the docs.

Removing the `array` type on the variable restores fully normal functionality.

src/StaticAnalyser.php
```diff
@@ -391,4 +390,30 @@ private function parseUseStatement(&$tokens, &$token, $parseContext)
...
+            $token = $this->nextToken($tokens, $parseContext);
+        }
+
+        if ($token === '?') { // nullable type
+            $nullable = true;
+            $token = $this->nextToken($tokens, $parseContext);
+        }
+
+        // drill down namespace segments to basename property type declaration
+        while ($token[0] === T_NS_SEPARATOR || $token[0] === T_STRING) {
```
darkbluesun, line 410: This should read 
```
while (in_array($token[0], [T_NS_SEPARATOR, T_STRING, T_ARRAY])) {
            if (in_array($token[0], [T_STRING, T_ARRAY])) {```

darkbluesun: although now I've thought about it, there's very little to be gained from the `array` type. The openApi document is going to need an 'items' property to be considered valid, and most codesniffer rules require you to have a var annotation for array types to specify what the items will be. I think it might be better if we just fix the handling of @var annotations for array types

darkbluesun: If we just add `T_ARRAY` to the `while` this will fix the problem reported by @SlvrEagle23 
If we add `T_ARRAY` to the if statement, it will override the `@var` annotation as you've designed it to do.
Actually why does native trump `@var`?

darkbluesun: Fixed in https://github.com/jpuck/swagger-php/pull/3

bfanger:
Rebased, squashed and merged into  0b63a79 (Includes PR 3 from  @darkbluesun )

Very nice work!

src/StaticAnalyser.php
jpuck: fabulous, thank you! sorry I was away from Github for a few days. so happy to see this open source carrying on without me 😄

PR #748, opened 2020-01-06
PHP 7.4 Preloading: define 'scan()' function only if not already defined

Baldinof:
Hi,

If the `functions.php` file is preloaded and then included at runtime, it triggers a `PHP Fatal error:  Cannot redeclare OpenApi\scan()`.

Constants are not preloaded, so this PR split constants and function definition to fix this issue.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -16,7 +16,9 @@ if (defined('OpenApi\UNDEFINED') === false) {
     define('OpenApi\UNDEFINED', '@OA\UNDEFINED🙈');
     define('OpenApi\Annotations\UNDEFINED', UNDEFINED);
     define('OpenApi\Processors\UNDEFINED', UNDEFINED);
+}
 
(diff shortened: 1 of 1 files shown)


michaljusiega:
Confirm same error with PHP 7.4.3 and Symfony 4.4 preloading.php file.

vsychov:
PR for fix same problem in 2.x branch - https://github.com/zircote/swagger-php/pull/758

DavidGarciaCat:
Just out of curiosity: is anyone merging this update and releasing a new tag so we can get it via `composer update`? I can see that it has been open for 4 months although there are a few approvals 🤔🤷‍♂️

Doqnach:
Unfortunately the owner of the repo has been MIA for many months, and the involvement of any contributors has been on the low side. This project desperately needs some more active contributors.

PR #758, opened 2020-02-27
PHP 7.4 Preloading: define 'scan()' function only if not already defined for 2.x branch

vsychov:
Hi guys,

same fixes to zircote#748 but for 2.x version


achasseux:
@zircote Do you have any idea when this will be merged  ?

DavidGarciaCat:
Just out of curiosity: is anyone merging this update and releasing a new tag so we can get it via composer update? I can see that it has been open for 4 months although there are a few approvals 🤔🤷‍♂️

This update fixes my problems with Symfony 4.4 and OPcache Preload, so 👍 for me

PR #761, opened 2020-03-09
Update Getting-started.md

polishExperiment:


uuf6429:
Hey, the "mistake" is intentional. Read the your fix again:
```
For example if you'd write `title="My first ...` swagger-php whould generate a notice with
                            ^^^^^
 "Unexpected field "title" for @SWG\Info(), expecting "title", ..."
                    ^^^^^                              ^^^^^
```

If the "title" was written, you wouldn't get the notice. This is just an example of how it handles mistakes.


On the other hand, `whould` is definitely a mistake, it should be `would`.

polishExperiment:
Ok, then I'm closing it

PR #773, opened 2020-05-08
Refactor tests

DerManoMann:
**Changes**
* Mirror the namespaces from `src/` in the `tests/` folder.
* Add some helper methods around fixtures (resolve, load)
* Clarify a couple comments and test names to what I believe they actually do
* Remove redundant dataProvider in favour of shared `OpenApiTestCase::allAnnotationClasses()`

**Motivation**
I've started looking at the current traits support and got really annoyed by the long list of test cases in the `tests` folder itself.
Working my  way backwards through the code I found it hard to see which library classes have corresponding test cases and which not.
Also, before going any further I wanted to see first if my style / thinking is appropriate before committing to more work...

Looking forward I plan on adding more tests first, in particular around the way the code uses `Context::$class` as that seems to be the root of (at least some) evil.
Only after that I'll move forward with actual code changes itself.

### DIFF
--- a/tests/Annotations/AbstractAnnotationTest.php
+++ b/tests/Annotations/AbstractAnnotationTest.php
@@ -4,7 +4,9 @@
  * @license Apache 2.0
  */
 
-namespace OpenApiTests;
+namespace OpenApiTests\Annotations;
+
+use OpenApiTests\OpenApiTestCase;
 
 class AbstractAnnotationTest extends OpenApiTestCase
 {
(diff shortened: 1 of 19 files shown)


.gitignore
```diff
@@ -5,4 +5,5 @@
 /openapi.json
 /openapi.yaml
 /docs/.vuepress/dist
-.phpunit.result.cache
\ No newline at end of file
+.phpunit.result.cache
+.idea
```
Doqnach, line 9: I would suggest keeping `.idea` in your own local global git-ignore.
It gets a mess if you have to start adding ignores for each and every IDE/editor for all contributors.

DerManoMann: Fair enough - habbit from working mostly on my own projects :)

Issue #750, opened 2020-01-10
How to use traits and OA annotations the proper way?


mroeling:
Reading some of the issues (#700 , #576 ), and still having some problems myself, I decided to setup a simple data model. Defined are the following models and traits:
* Dog extends Pet, no additional properties
* Pet uses Food trait and Fur trait, and has the property name
* Fur trait uses FurColor trait and has the property fur
* FurColor has the proporty color

Note: because of the mentioned problem of separate namespaces, I applied that as well, since our production environment uses a similar approach. 

As I will mention in the result section I ran the code on both the 3.0.3 version and the dev-trait-reference-fix branch. The result was then checked through https://editor.swagger.io/.

### Files

Dog.php
```php
<?php
use OpenApi\Annotations as OA;

namespace animals\object;

/**
 * Class Dog
 * @package animals\object
 * @OA\Schema(schema="Dog",
 *     allOf={
 *     @OA\Schema(ref="#/components/schemas/Pet"),
 *     }
 * )
 */
class Dog extends Pet
{
}
```

Pet.php
```php
<?php
use OpenApi\Annotations as OA;

namespace animals\object;

use animals\traits\FoodTrait;
use animals\traits\FurTrait;

/**
 * Class Pet
 * @package animals\object
 * @OA\Schema(schema="Pet",
 *     allOf={
 *         @OA\Schema(ref="#/components/schemas/HasFood"),
 *         @OA\Schema(ref="#/components/schemas/Fur")
 *     }, required={"name"}
 *     )
 */
class Pet
{
    use FoodTrait;
    use FurTrait;

    /**
     * @OA\Property(
     *     type="string"
     * )
     * @var string
     */
    protected $name;
}
```

FurTrait.php
```php
<?php
use OpenApi\Annotations as OA;

namespace animals\traits;

/**
 * Trait FurTrait
 * @package animals\traits
 * @OA\Schema(schema="Fur", allOf={
 *         @OA\Schema(ref="#/components/schemas/FurColor")
 *     }
 * )
 */
trait FurTrait
{
    use FurColorTrait;
    /**
     * @OA\Property(type="string")
     * @var string
     */
    public $fur;
}
```

FurColor.php
```php
<?php
use OpenApi\Annotations as OA;

namespace animals\traits;

/**
 * Trait FurColorTrait
 * @package animals\traits
 * @OA\Schema(schema="FurColor")
 */
trait FurColorTrait
{
    /**
     * @OA\Property(type="string")
[...]

PR #774, opened 2020-05-13
Traits cleanup

DerManoMann:
# WIP #

Some cleanup and fixes for trait handling (as it was intended IMHO).

* Deal with nested traits (traits using other traits)
* Fix operationId for traits and interfaces
* Fix some quirk in `fullyQualifiedName()` that would modify itself and break things further down the track
* Use correct var for trait property names in `ImportTraits` processor
* Added traits example as reference

The result of these changes is that **all** trait properties are merged into the class using the trait. A side effect is that the traits are still showing up as schema in the output, but are basically empty.

I suppose the alternative would be to **not** merge properties and use `allOf` on classes that use traits. Not sure yet what is better (or more `correct`), though.

Some discussion would be useful before processing any further....


mroeling:
Great effort! I think it fits the expected result for #750.

Looking forward to see the use of references as well. Something like: Product extends SimpleProduct, with extra traits.

DerManoMann:
> Looking forward to see the use of references as well. Something like: Product extends SimpleProduct, with extra traits.

Hmm, I would have thought inheriting from parent classes should work as. If you referring to the use of `allOf`, that is an alternative approach but not something that would happen at the same time as merging properties?

DerManoMann:
So, a few  more questions to discuss...
1. Currently the validation will log a warning due to a [duplicate property](https://github.com/DerManoMann/swagger-php/blob/traits-example/Examples/using-traits/Product.php#L24).
Technically, this is valid and the class property/annotation should win (and it does). So, should the warning be supressed if the property comes from a trait? Maybe I am just not sure what to do with those warnings...  
1. Looking at the output the traits are all empty (merged) and look funny in swagger-ui. Should they be removed as nothing is referencing them either?
1. I might add a separate processor for importing traits as `allOf` refs instead of merging properties. Maybe people could then chose one or the other?

mroeling:
In general: IMHO no functionality should be removed when parsing @OA into (yml) output.

1) Although it might be possible to double define a property, IMHO it's more like bad design... So I would prefer to have the warning in these cases. Since it is possible it doesn't have to be a failure, but a warning should fit the situation.

2) and 3) OA does support references, both for the API calls and the allOf implementation. So that's one of the reasons to prefer to have a (separate) processor. Another reason is rather personal. Our company not only uses the generated YML (or json, since we use 2.x because of the lack of support yet in 3.x) for the API calls, but also as base for creating models and interfaces in Typescript, using the swagger-schema-official and swagger-parser packages.
From the OA documentation: 
> An OpenAPI definition can then be used by documentation generation tools to display the API, **code generation tools to generate servers and clients** in various programming languages, testing tools, and many other use cases.
This would not be possible when traits are deleted.

bfanger:
@zircote Could you add @DerManoMann as collaborator?

tests/ExamplesTest.php
```diff
@@ -31,15 +36,16 @@ public function testExample($example, $output)
...
-            ['misc', 'misc.json'],
-            ['openapi-spec', 'openapi-spec.json'],
-            ['petstore.swagger.io', 'petstore.swagger.io.json'],
-            ['petstore-3.0', 'petstore-3.0.json'],
-            ['swagger-spec/petstore', 'petstore.json'],
-            ['swagger-spec/petstore-simple', 'petstore-simple.json'],
-            ['swagger-spec/petstore-with-external-docs', 'petstore-with-external-docs.json'],
-            ['using-refs', 'using-refs.json'],
-            ['example-object', 'example-object.json'],
+//            ['misc', 'misc.json', []],
```
bfanger, line 39: Is it possible to add a custom command line option to phpunit? Something like:
`phpunit --swagger-example using-traits` 
Right now the other examples are excluded from running in travisCI and easy to forget to re-enable when te PR is ready.

DerManoMann: Yeah, that was an oversight from doing manual testing. I'll restore that once we are happy with the changes.
I guess I could user `--filter using-traits` but then it gets complicated when trying to debug in the IDE...

DerManoMann:
@mroeling thanks for the comments. 

I think I agree with you on all points and unless there is more opposition I'll continue working on a new processor that merges traits as `allOf` refs.

Also, I think we just have to separate the annotation processing from the PHP level rules about conflicts and overriding of properties, etc.

That means it totally makes sense to keep all the validation warnings in place and its the authors responsibility to keep things sane :)

mroeling:
@DerManoMann Great!

DerManoMann:
Closing this as it was never intended to be merged in this form. All relevant code changes are covered by other PRs.

PR #775, opened 2020-05-14
Add missing use statements to MergeXmlContent

DerManoMann:
Adds missing use statements and corresponding test(s).
Also adds `ext-json` as requirement to `composer.json`.

@bfanger sorry for adding nw PRs now that you've cleaned up so nicely.
Feels to me that changes to the repo have been a bit disjointed so there are all sorts of little mismatches. [Also, code coverage is a bit low, IMHO :)]

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,10 +26,12 @@
     }
   ],
   "config": {
-    "bin-dir": "bin"
+    "bin-dir": "bin",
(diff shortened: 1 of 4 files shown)


bfanger:
Most PRs add code without unittests, lowering coverage.

Nice to see a PR increasing coverage.👍

PR #776, opened 2020-05-14
Move examples output files next to the example code

DerManoMann:
To me it feels more natural to have the generated openapi definitions next to the sources.

This PR also renames the markup file in the examples folder so it is automatically displayed when the folder is browsed on GH.


bfanger:
I agree, except that the files should use the yaml format (preferred since openapi 3) instead of json.

DerManoMann:
> I agree, except that the files should use the yaml format (preferred since openapi 3) instead of json.

Oh, I must have missed your comment :/ Happened to look at the PR today to rebase... and then went down the YAML rabbit hole...
On the plus side things are now again a bit nicer :)

bfanger:
Amazing work! Thanks again!

PR #777, opened 2020-05-15
Change reference to _context variable to $parent

piratadelfuturo:
I was getting an error about error reporting from this mis referenced variable.


DerManoMann:
Cool. Could you please extend the PR and fix the same line in `MergeXmlContent` too?

Also, if possible it would be great to add a simple testcase. Thanks.

bfanger:
Thanks for reporting, closed in favour of #800

PR #780, opened 2020-05-16
Improve `Context::detect()`

DerManoMann:
Add support for detecting used traits to context detect.


DerManoMann:
I think at this stage using reflection is not in the right spirit of things...

Issue #663, opened 2019-03-23
Can't generate Query Parameters automatically like I can request body


JoeyOverby:
I am writing an OpenAPI spec and trying to generate my possible query parameters automatically (using swagger-php) from the annotations for a request route/path. I know that I can just type out all possible parameter options for each route, but I really need to be able to generate the possible parameters from the properties of a class automatically using annotations like I can do for the request body. (We will have tons of classes/paths and keeping it up to date most likely won't happen unless they are generated like the request body/JsonContent can be. Is this possible with swagger-php or even OpenAPI in general?

I got this to work with the put and the request body, but how do I do it for a get request that still uses the properties of the class?

I can do this for the request body:

```php
    /**
     * @return Response
     *
     * * @OA\Put(
     *     path="/persons",
     *     tags={"Person"},
     *     @OA\RequestBody(
     *          request="person",
     *          required=false,
     *          description="Optional Request Parameters for Querying",
     *          @OA\JsonContent(ref="#/components/schemas/Person")
     *      ),
     *     @OA\Response(
     *          response="200",
     *          description="Returns matching Person Object",
     *          @OA\JsonContent(
     *              type="array",
     *              @OA\Items(ref="#/components/schemas/Person")
     *          )
     *     )
     * )
     */
```

Writing out each parameter for 30+ classes will not be maintainable:

```php
     /** @OA\Get(
     *     path="/events",
     *     tags={"Events"},
     *     @OA\Parameter(
     *          name="eventID",
     *          in="query",
     *          required=false,
     *          description="The event ID specific to this event",
     *          @OA\Schema(
     *              type="string"
     *          ),
     *     ),
    *
   * ....etc
```

bfanger:
Write the parameters outside the `@OA\Get` 
```
    /**
     * @OA\Parameter(
    *          parameter="event_id",
     *          name="eventID",
     *          in="query",
     *          required=false,
     *          description="The event ID specific to this event",
     *          @OA\Schema(
     *              type="string"
     *          )
     *     )
     */

And use a ref inside the `@OA\Get` to reference that parameter.

   /** @OA\Get(
     *     path="/events",
     *     tags={"Events"},
     *     @OA\Parameter(ref="#/components/parameters/event_id")
     *      ...
```

This allows you to document the parameters and re-use them multiple times. This is generally do-able. 

You can also look into writing a swagger-php plugin, a processor that generates parameter annotations based on the class information.

JoeyOverby:
[...]

PR #783, opened 2020-05-21
Add simple example of using a custom processor

DerManoMann:
Custom processor that injects parameters into operations based on a
vendor extension annotation.

Showcases a solution for #663

### DIFF
--- a/Examples/schema-query-parameter-processor/SchemaQueryParameter.php
+++ b/Examples/schema-query-parameter-processor/SchemaQueryParameter.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace SchemaQueryParameterProcessor;
+
+use OpenApi\Analysis;
+use OpenApi\Annotations\Components;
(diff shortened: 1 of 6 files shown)


jeandonaldroselin:
I try to make this work with my Symfony 6 project, but I cannot.
Maybe I missed a Step :

Is the file scan.php required to make it work ?
If yes where we are suppose to import + use it ?

Thank you in advance

DerManoMann:
The `scan.php` is just there to play around. The point is to add the processor to your swagger-php setup. Depending on whether you are using swagger-php directly or via bundle/extension there might be different ways to configure this.

jeandonaldroselin:
> The `scan.php` is just there to play around. The point is to add the processor to your swagger-php setup. Depending on whether you are using swagger-php directly or via bundle/extension there might be different ways to configure this.

Thank you for the reply, I use the Symfony 7 bundle so I will check the provided options for this case.

PR #786, opened 2020-05-25
Fix context variable

cristian-chiric:


Doqnach:
This looks to fix the same thing as https://github.com/zircote/swagger-php/pull/777/files#diff-fc981ecc7c4b6ef671e291895ecb7f88 does, but referencing a different context... 
If I understand the code correctly I would say that PR #777 looks correct since the surrounding code has `$parent` as its context.

pinging @piratadelfuturo and @DerManoMann

bfanger:
Could you provide examples when the error occurs? (preferably in a unittest) 
That would help us choose which context is the right one, or even change the context based on the error condition.

DerManoMann:
Same as #777

bfanger:
Thanks for reporting, closed in favour of #800

PR #792, opened 2020-05-29
Fixed version doctrine/annotations in require dependencies

AnnaBuzian:
A bug has been fixed for version php 7.4 in version 1.7 of doctrine/annotations. If you install the zircote/swagger-php 3.0.4 now, then doctrine/annotations won't be updated and you will get an error: "Trying to access array offset on value of type null".

Link to package release:
https://github.com/doctrine/annotations/releases/tag/v1.7.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
   "require": {
     "php": ">=7.2",
     "ext-json": "*",
-    "doctrine/annotations": "*",
+    "doctrine/annotations": ">=1.7",
     "symfony/finder": ">=2.2",
     "symfony/yaml": ">=3.3"
   },


bfanger:
Thanks, but I won't merge this pul request.

Version pinning in a project composer.json is good practice, but in a library we must be more liberal. Especially with a package that other packages also depend on, this is to make sure composer can find a suitable version.

Sadly, composer is not smart enough to to detect which packages no longer work on newer php versions (design flaw?). 
But in this case, an incompatible sub-package: `doctrine/annotations`, solving the issue is easy by also upgrading that to 1.7.* or higher in the composer.json **of the project**

Thanks again for reporting, this will make fixing the issue for others easier.

PR #819, opened 2020-07-22
Code style

DerManoMann:
Switch to [php-cs-fixer](https://github.com/FriendsOfPhp/PHP-CS-Fixer) for code style checks.
* there are a lot more options
* it allows to fix things automatically

This branch also includes some namespace cleanup for tests and fixtures that used to annoy PHPStorm...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /openapi.yaml
 /docs/.vuepress/dist
 .phpunit.result.cache
+/.php_cs.cache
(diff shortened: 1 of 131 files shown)


DerManoMann:
Mostly phpdoc changes plus some cleanup in the `use` statements.
If noone objects I'll merge this in the next few days...

thePanz:
I am curious about the decision to use `'concat_space' => ['spacing' => 'one'],`.
Do you have any input about diverging from the default value there?

DerManoMann:
> I am curious about the decision to use `'concat_space' => ['spacing' => 'one'],`.
> Do you have any input about diverging from the default value there?

Nothing scientific - I suppose it was a matter of taste and what I use elsewhere. Having looked at what others do I guess we can change it to `true` to go with the default as long as it is applied consistently.

PR #830, opened 2020-08-18
Add support for new PHP8 T_NAME_xxx token

DerManoMann:
PHP8 related changes to namespace parsing.

https://wiki.php.net/rfc/namespaced_names_as_token


DerManoMann:
Bare minimum to make tests pass on PHP8 beta1

PR #840, opened 2020-09-02
Typehint and phpdoc improvements based on phpstan

DerManoMann:
Fix low hanging fruit from running `phpstan` with level 2 checking.

Also adds a basic `psalm` config and a new composer run-script target ('analyse') to execute both tools.

Tool dependencies are intentially not added to avoid unnecessary bloat and download time in travis.

This is intended as a purely manual tool for regression checking and a help to iteratively improve the codebase.


DerManoMann:
@bfanger Happy for you to merge this if you think this is ok. Impact is pretty minor but it is nice to have some baseline commands in the future...

bfanger:
Yes, i assumed you where still working on addional changes.

Issue #733, opened 2019-11-24
How to use class doc-block comment as the Schema description?


scipilot:
It's great that properties pick up their `description` from the doc-block comment.

However it seems that the doc-block comment for a model as a Schema doesn't do the same. So I have to duplicate the description in the annotations. I've looked through the documentation and also tried to follow the code, but cannot see if I've done something wrong, or this isn't supported. I can't see any closed or open issues either, I'm surprised no one else has requested it.

Am I missing something or is this how it's supposed to work?

e.g. I have many models, to which I simply added `@OA\Property()` annotations to, and they look pretty good in the swagger docs with their existing comments.

But the main schema description doesn't get scraped from the class doc-block, so I have to copy it in:

```
/**
 * News articles are announcements from the test admin team.
 * 
 * @OA\Schema(description="News articles are announcements from the test admin team.")
 */
class NewsArticle {

	/**
	 * Article Title
	 * @var string
	 * @OA\Property()
	 */
	public $title;

	/**
	 * Main article body content.
	 * @var string
	 * @OA\Property()
	 */
	public $body;

	/**
	 * Publish/sort date.
	 * Enables the list of articles to be presented on the site in date-order.
	 * 
	 * @var \DateTime
	 * @OA\Property()
	 */
	public $published;
}
```

Is there a way to automate it and keep it DRY?

bfanger:
This feature will be part of the next release

PR #844, opened 2020-09-05
feat: Infers summary and description from plain text in commentblock

bfanger:
Implements #733

In semver is this a feature or a breaking change?
I expect lots of additional descriptions appearing in api docs.


DerManoMann:
i do not think this breaks anything. it's more of a feature or improvement as it doesn't break BC and doesn't change anything in case the properties are explicitly set

bfanger:
When we expect it to "break" with a php error (ex. renaming a public function) we'll bump the mayor version. but when OpenApi\scan() with the same input generates different output.that's a bump in minor of patch version depending on the change.

PR #850, opened 2020-09-14
Combine merge/inherit interfaces/traits processors

### LINKED ISSUE #847: Interface schema can be null
With the fix for #842, the following code was introduced.
This breaks if the interfaceSchema is null.

https://github.com/zircote/swagger-php/blob/9d9f038a97d02f043abbacaa6248088a4803d0e3/src/Processors/InheritInterfaces.php#L23-L25

In the InheritTraits class this was correctly implemented, see
https://github.com/zircote/swagger-php/blob/9d9f038a97d02f043abbacaa6248088a4803d0e3/src/Processors/InheritTraits.php#L25-L27

DerManoMann:
Reduce to just a single processor for each interfaces and traits.

Also update the `AugmentSchemas` processor to move properties into a nested `allOf` schema if both `properties` and `allOf` are populated.

Fixes #847


bfanger:
👍 kudus for making testExamples() all use the default processors.

DerManoMann:
yes, much nicer this way

DerManoMann:
@bfanger just double checking - you think this is good to merge?

PR #852, opened 2020-09-15
Fix constant resolution in files with namespaces

### LINKED ISSUE #851: Resolution Failures in files with Namespaces
Referencing constants defined in the same file as the Annotation itself yields resolution errors. For example, consider running `StaticAnalyzer::fromFile()` on a file with the following content:

```php
<?php
namespace Example;

/**
 * @Schema(example=MyClass::CONSTANT)
 */
class MyClass
{
    const CONSTANT = 'some value';
}
```

Processing fails with an error of:
`Doctrine\Common\Annotations\AnnotationException: [Semantical Error] Couldn't find constant MyClass::CONSTANT`

The error seems to lie within the namespace processing. Currently, the `StaticAnalyzer` does not pass information about the namespace of the processed file (or token stream) to the `DocParser` used, such that constants of classes defined in the same file are referenced as if belonging to no namespace at all, which leads to the Doctrine Annotation Reader not finding the constant's definition.

jabrandes:
Presumably closes #851 by setting the special import key `__NAMESPACE__` that is considered by the `DocParser` during constant resolution.

The problem specified in #851 is tested for by a modification of `Parser/User.php` fixture, in which a class constant is referenced to provide an example value in `@Schema` class annotation.

### DIFF
--- a/src/StaticAnalyser.php
+++ b/src/StaticAnalyser.php
@@ -349,6 +349,8 @@ class StaticAnalyser
 
             if ($token[0] === T_NAMESPACE) {
                 $parseContext->namespace = $this->parseNamespace($tokens, $token, $parseContext);
+                $imports['__NAMESPACE__'] = $parseContext->namespace;
+                $analyser->docParser->setImports($imports);
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


DerManoMann:
Looking good. Could you perhaps add a small test to `tests/StaticAnalyserTest` to verify that the value is actually there?

Something like
```
    public function testNamespacedConst()
    {
        $analysis = $this->analysisFromFixtures('Parser/User.php');
        $schemas = $analysis->getAnnotationsOfType(Schema::class, true);
        $this->assertCount(1, $schemas);
        $this->assertEquals('value', $schemas[0]->example);
    }
```

jabrandes:
@DerManoMann, I just added your test case since I do not think there is much more to test for than what you already proposed. I only made the slight modification to compare against `User::CONSTANT` directly.

Issue #839, opened 2020-09-01
Apply semantic versioning


andi-wr:
I have `"zircote/swagger-php": "3.0.x"` as a dependency in my `composer.json`. Today my build broke due to a strange error while parsing files of my library with `nikic/php-parser`. After hours of debugging I found out that the error occurs with 'nikic/php-parser >= 4.7.0' in combination with `zircote/swagger-php >= 3.0.5`. The bug is already fixed in `d61e2a4d9f32d3cf70f1e6c84ee181ffb270fae1` but there is still no version tag `3.0.6` available.

But independent of the availability of `3.0.6` - is it possible to apply semantic versioning in this library for future releases? If you follow those rules, then `3.0.x` means that I will only receive patches (bugfixes or security issues) when updating my dependencies, but no new features which can lead to side effects with other third party dependencies or my own code. In my opinion, "Add support for new PHP8 T_NAME_xxx token" is clearly a new feature and should result at least in a new minor version.

Thank you!

DerManoMann:
First of all, sorry for breaking things, this was definitely not intended.

I am with you on semver (https://github.com/zircote/swagger-php/pull/782#issuecomment-636413726) but my hands are tied on that one. I only have commit rights but cannot update packagist or create releases. Seeing that @zircote seems to be MIA most of the time and @bfanger is busy with other stuff and slowly moving away from this project options are limited.

Then again, I am not sure this would have been a feature change. Ideally nothing would have changed in behavior other than the same features would work with PHP8. Maybe a discussion for another day since I do not have any control over this.

bfanger:
This project doesn't use semanic versioning yet.

The mayor version number has always matched the spec version. 

 i didn't see changes in our public api (tests passed) some features would just started behaving correctly.

Im not against sevmver, but it's just hard to do, there is a lot that could be considered breaking and bumping the mayor version for every (small) breaking change is also a hassle for adoption.

But looks @DerManoMann has some ideas on how to apply semver 👍

michaljusiega:
See https://github.com/nikic/PHP-Parser/pull/706#issuecomment-696002603 - we must add:

```
if (!defined('T_ATTRIBUTE')) {
    define('T_ATTRIBUTE', -6);
}
```

in `functions.php`

PR #857, opened 2020-09-21
Fix T_ATTRIBUTE

michaljusiega:
So, this PR fixes issue similar to https://github.com/zircote/swagger-php/issues/839, https://github.com/zircote/swagger-php/pull/837, https://github.com/nikic/PHP-Parser/pull/706 but after bumping to `nikic/php-parser:4.10.0` and `zircote/swagger-php:3.1.0` the error occurs with the token `T_ATTRIBUTE`. See comment (https://github.com/nikic/PHP-Parser/pull/706#issuecomment-696002603)

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -25,6 +25,9 @@ if (!defined('T_NAME_QUALIFIED')) {
 if (!defined('T_NAME_FULLY_QUALIFIED')) {
     define('T_NAME_FULLY_QUALIFIED', -5);
 }
+if (!defined('T_ATTRIBUTE')) {
+    define('T_ATTRIBUTE', -6);
+}
 
 if (function_exists('OpenApi\scan') === false) {
     /**


DerManoMann:
I can see where this is going but I am a bit confused - the swagger-php tests pass with `nikic/php-parser:4.10.0` and one of my projects also still works with the latest versions of both.
Also, with the other two PHP8 token things are a bit different as we do use those in the codebase. `T_ATTRIBUTE` must be a new one but we do not use that at all, so adding it in this project seems the wrong way to me.

michaljusiega:
Unfortunately, in my case, adding a token helps to fix the error.

DerManoMann:
Fair enough. my point then is that you could add that code to anu of your 3rd party libraries or even your own codebase. IMHO this fix is not a swagger-php issue.

michaljusiega:
I just had to add manually to get the problem gone. Well, if you think so, you can close it. :)

DerManoMann:
Closing for now. swagger-php is not using `T_ATTRIBUTE` at all so this looks like an issue in  `nikic/PHP-Parser`

PR #859, opened 2020-09-22
Bump doctrine deps and cleanup some cruft

DerManoMann:
Bump doctrine dependency to 1.7 to make things work in 7.3/7.4 even with the lowest possible dependencies 
(`composer --prefer-lowest`)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
   "require": {
     "php": ">=7.2",
     "ext-json": "*",
-    "doctrine/annotations": "*",
+    "doctrine/annotations": "^1.7",
     "symfony/finder": ">=2.2",
     "symfony/yaml": ">=3.3"
   },
--- a/src/Analyser.php
+++ b/src/Analyser.php
@@ -6,12 +6,11 @@
 
 namespace OpenApi;
 
-use Exception;
+use Doctrine\Common\Annotations\AnnotationRegistry;
+use Doctrine\Common\Annotations\DocParser;
 
-if (class_exists('Doctrine\Common\Annotations\AnnotationRegistry', true)) {
-    // Using doctrine/annotation 1.x
-    // Load all whitelisted annotations
-    \Doctrine\Common\Annotations\AnnotationRegistry::registerLoader(
+if (class_exists(AnnotationRegistry::class, true)) {
+    AnnotationRegistry::registerLoader(
         function ($class) {
             if (Analyser::$whitelist === false) {
                 $whitelist = ['OpenApi\Annotations\\'];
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -37,7 +37,7 @@
   "require": {
     "php": ">=7.2",
     "ext-json": "*",
-    "doctrine/annotations": "*",
+    "doctrine/annotations": "^1.7",
```
bfanger, line 40: This would block interop with frameworks that have set a 1.6 or lower as dep.
And swagger-php doesn't require 1.7 i think.

src/Analyser.php
```diff
@@ -6,12 +6,11 @@
 
 namespace OpenApi;
 
-use Exception;
```
bfanger, line 9: Is removing  `use Exception` but adding a slash a common convention / better? (i genuinely don't know)

bfanger:
Related  #437 and #792

Sadly version management in composer (and npm) is broken as you'll need to specify the future versions of the deps your release is compatible with, which works in theory with semver, but in practice is impossible.

src/Analyser.php
Doqnach: As long as you don't forget the backslash, neither is actually better I believe. Leaving out the backslash does make it marginally slower because it would have to search both the current namespace and the global namespace...

It depends on the code-style really. Personally I always import (`use`) everything.

composer.json
DerManoMann: It breaks using PHP 7.4 and running composer with `--prefer-lowest`.
doctrine 1.7 was released August 2019 - I really cannot imagine anyone really relying on older versions

src/Analyser.php
DerManoMann: It is a style/preference question. I am leaning towards the Symfony styleguide for the most part and that is how its used elsehwere in the codebase that I have touched.

DerManoMann: @bfanger  Reading through this again - when would you be ok with bumping this then? Active support for PHP 7.3 runs out next month and for PHP 7.4 in a year. After that, at the latest, 1.7 would be the min required version IMO...
In practice I would be surprised if this change would break things for anyone.

bfanger: What about  ">=1.7" ? Our usage of doctrine could be compatible with 2.x and up.

Probably a good idea to and bump "php": ">=7.4" and drop 7.3 and lower from travis.

DerManoMann: Back to the semver question, eh? I do not agree that "could be compatible" is good enough and who knows if/when that would be. From my experience, once there is a 2.0 we can test and if it is compatible bump deps then.
Noone expects a library to support a new major version of another library from day one, IMHO.
In fact, in that respect the current `"php": ">=7.2"` was wrong for quite a while because PHP8 was broken...

bfanger: It's a broken system, but better than no system at al. If we choose ^1.7 will Composer will select the current swaggr-php version? (as * is compatible with 2). But that aside going for a more safe settings seems like a good. path forward. Merge approved

PR #860, opened 2020-09-24
Validate that @OA\Items parent has type "array"

### LINKED ISSUE #846: Having properties inside items in a schema does not render properly in Swagger UI
Hi, I have a schema that has several properties which can be seen below and one of them is called `data` its type is an object. I tried to add all object properties inside @OA\Items and generate the spec in json. Then I check it inside swagger web UI and `data` property was not render properly and was showing an empty object. Then I took the properties outside of @OA\Items and it worked. By the way I uploaded my spec in swaggerhub editor and it did not show any error messages, also in swaggerhub UI property `data` was empty. Is the first attempt using @OA\Items a valid structure, if not shouldn't swagger cmd that generates the spec `php vendor/bin/openapi `  give me an error instead.

Before:
```
@OA\Schema(
    schema="MySchema",
    @OA\Property(property="status", type="integer"),
    @OA\Property(
        property="data",
        type="object",
        @OA\Items(
            @OA\Property(property="message", type="string"),
            @OA\Property(property="valid", type="boolean")
	   )
	)
)


Example Value
Schema
{
  "status": 0,
  "data": {}
}
````

After:
```
@OA\Schema(
    schema="MySchema",
    @OA\Property(property="status", type="integer"),
    @OA\Property(
        property="data",
        type="object",
            @OA\Property(property="message", type="string"),
            @OA\Property(property="valid", type="boolean")
     )
)


Example Value
Schema
{
  "status": 0,
  "data": {
    "message": "string",
    "valid": true
    }
  }
}
```

### PR DESCRIPTION
**Pull request by DerManoMann**
Adds validation the the `Items` annotation to check that the parent is of type `"array"`.

Also includes some cleanup of defunct tests and cleanup...

Fixes #846.

### DIFF
--- a/src/Annotations/Items.php
+++ b/src/Annotations/Items.php
@@ -46,18 +46,13 @@ class Items extends Schema
         if (in_array($this, $skip, true)) {
             return true;
         }
+
         $valid = parent::validate($parents, $skip);
-        if ($this->ref === UNDEFINED) {
-            $parent = end($parents);
-            if (is_object($parent) && ($parent instanceof Parameter && $parent->in !== 'body' || $parent instanceof Header)) {
-                // This is a "Items Object" https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#items-object
-                // A limited subset of JSON-Schema's items object.
-                $allowedTypes = ['string', 'number', 'integer', 'boolean', 'array'];
-                if (in_array($this->type, $allowedTypes) === false) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Annotations/Items.php [hunk 1]
```diff
@@ -46,18 +46,13 @@ public function validate($parents = [], $skip = [], $ref = '')
         if (in_array($this, $skip, true)) {
             return true;
         }
+
         $valid = parent::validate($parents, $skip);
-        if ($this->ref === UNDEFINED) {
```
DerManoMann, line 50: The associated test was effectively disabled...

PR #873, opened 2020-10-29
Add ability to include multiple other directories

thewoods96:
### Description
A small change that allows 1-* additional directories to be specified using the `--include` param. In addition to the `--exclude` option this offers a granular way of specifying exactly what should be scanned.

#### Background
The project that we are currently configuring swagger-php within includes an API client package for a downstream API, all the models are defined there and we'd like to create schemas and reference them within our API. 

We came across the problem that we couldn't just include our package without the entire contents of the vendor folder, therefore we built this option so we can simply specify our package to include. We figured this could be useful for others in similar situations also!

### DIFF
--- a/bin/openapi
+++ b/bin/openapi
@@ -13,6 +13,7 @@ error_reporting(E_ALL);
 $options = [
     'output' => false,
     'format' => 'auto',
+    'include' => [],
     'exclude' => [],
     'pattern' => '*.php',
     'bootstrap' => false,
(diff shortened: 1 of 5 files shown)


DerManoMann:
Hmm.

Done well and even with tests :) I just can't immediately see what the difference would be to providing multiple paths on the CL?

DerManoMann:
Closing for now. Any future improvements should be based on the new `Generator` class

PR #881, opened 2020-11-25
Update Store.php: fix minimum error

wangyongtao:
fix error: Semantic error at paths./store/order/{orderId}.get.parameters.0.schema.minimum: 'minimum' must be lower value than 'maximum'


DerManoMann:
Thanks @wangyongtao .

PR #882, opened 2020-11-25
Fix PHP 8 compatibility in 2.x

Tobion:
My library fails with php 8 using swagger 2.x, see https://github.com/Tobion/OpenAPI-Symfony-Routing/pull/7
This is due to the namespace not being parsed correctly.

This PR fixes PHP 8 compatibility.
It would be very hard to test with php 5.6 until php 8 due to phpunit bc breaks and incompatibilities, e.g. phpunit <9 will not support php 8. so just droppping unsupported php versions is the best choice which was also done in 3.x (master).

This backports #830, #837, 29e913a, parts of 3f1cee9 to fix php 8 errors and tests.


Tobion:
@DerManoMann tests pass. I would appreciate if you tag this as 2.1 release.

composer.json
```diff
@@ -33,13 +33,13 @@
...
-        "zendframework/zend-form": "<2.8",
         "squizlabs/php_codesniffer": ">=2.7",
-        "phpunit/phpunit": ">=4.8.35 <=5.6"
+        "phpunit/phpunit": "^8 || ^9"
     },
     "autoload-dev": {
         "psr-4": {
-            "SwaggerTests\\": "tests/"
+            "SwaggerTests\\": "tests/",
+            "AnotherNamespace\\": "tests/Fixtures/AnotherNamespace"
```
DerManoMann, line 42: Noit sure if this is really needed seeing that the fixtures do not properly use PSR-4 namespaces.

DerManoMann (approved):
Nice work - thanks.

composer.json
Tobion: The fixtures use psr-4 namespace autoloading corrctly. Just with strange namespaces (\SwaggerTests, \AnotherNamespace). So I think this is needed to load the annotation examples.

DerManoMann: Hmm, I would argue that
```
namespace SwaggerFixtures;
```
is not a correct namespace according to PSR-0/PSR-4 as it does not match the folder name at all.

Nevertheless, that's not really important for the fixtures so all is good.

DerManoMann:
@bfanger @zircote I agree with @Tobion that a new tag/release for this would be nice.... As suggested, `2.1.0` seems like the right thing

Tobion:
@DerManoMann can't you create the tag?

DerManoMann:
I'll double check - looks like I can... I always mix up GH and packagist permissions :/

Tobion:
@DerManoMann thanks for the tag. Unfortunately the release is not visible to packagist: https://packagist.org/packages/zircote/swagger-php. Can you trigger a refresh?
And also enable auto-update: https://packagist.org/about#how-to-update-packages

DerManoMann:
damn, got distracted! autoupdate is definitely something I cannot configure

PR #884, opened 2020-12-12
Changed the Util finder method to force a provided instance of Finder…

snake14:
I used the option of passing an instance of Finder to the Util::finder method, since it made selecting multiple directories easy.  It took me a while to figure out why StaticAnalyser->fromFile(...) kept throwing an error because file_get_contents() was being executed on a directory.  To save other developers the time of figuring out that they have to set the Finder to only include paths for files, I figured that I would change Util::finder a little to enforce that, since the scan function is only expecting files anyway. I created a unit test to confirm that my change had the desired effect.

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -68,7 +68,8 @@ class Util
     public static function finder($directory, $exclude = null, $pattern = null): Finder
     {
         if ($directory instanceof Finder) {
-            return $directory;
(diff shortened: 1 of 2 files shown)


DerManoMann:
I guess one alternative would be to use a file extension mask?

PR #886, opened 2020-12-14
Github action CI migration

DerManoMann:
Looking at news around travis-ci and what other projecs are doing it looks like it is time to move on from travis :/

Gives better performance and a nicer build matrix testing each PHP version with lowest and highest composer deps.


DerManoMann:
@zircote @bfanger once merged we'd need to remove the travis check/integration on github. Not something I can do I think....

bfanger:
The Settings and toggle buttons are disabled in my  travis controlpanel. Should deleting the travis.yaml be a part of this PR?

DerManoMann:
Oh, well, we'll just have to wait and see then. Yes, that should be removed, ta.

DerManoMann:
> The Settings and toggle buttons are disabled in my travis controlpanel. 

@bfanger Hmm, disabling the webhook in github should be enough. Other than that this should be good to merge.
In the worst case we'll have to add an empty/noop `.travis.yml` file to make everyone happy.

PR #887, opened 2020-12-15
Don't crash when ref is null

hubertnnn:
If ref is set to null, then generating documentation will cause an exception: `substr() expects parameter 1 to be string, null given` on line 400
This happens for example in `nelmio/api-doc-bundle` when you ref from annotation defined documentation to one stored in config files.

This change will fix the problem by handling null the same as undefined values.

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -396,7 +396,7 @@ abstract class AbstractAnnotation implements \JsonSerializable
                 }
             }
         }
-        if (property_exists($this, 'ref') && $this->ref !== UNDEFINED) {
+        if (property_exists($this, 'ref') && $this->ref !== UNDEFINED && $this->ref !== null) {
             if (substr($this->ref, 0, 2) === '#/' && count($parents) > 0 && $parents[0] instanceof OpenApi) {
                 // Internal reference
                 try {


Doqnach:
@hubertnnn could you give an (as complete as possible) example when this would occur?

hubertnnn:
Ok, from more research I found it was a mistake on my part.

I had something like this in the yaml:
```
        Product:
          type: object
          properties:
            price:
              $ref: #/components/schemas/Price
            ...
```
that turned into comment instead of string.

But I still think, that the core problem still exists. 
If I would like to explicitly unset a ref (for later merging or any other reason) or if I will explicitly set it to null,
then it should not cause an exception. 

Especially since the code where this is located is responsible for detecting misconfigurations and logging them.
Even if we consider `$ref` set to `null` as a misconfiguration, having one type of misconfiguration causing a message to be logged and another causing an exception is at least confusing.

DerManoMann:
I am happy to merge this, but I struggle to understand the '#' issue. Modifying the YAML should not affect the code as swagger-php does not load the yaml if I am not mistaken.
If you programmatically change the `$ref` then things are different and you are (at least partly) responsible for the consequences :+1: 
However, I agree with your reasoning about handling unexpected `null` values...

Issue #880, opened 2020-11-25
Conflict with another library's annotations


rydan313:
I'm using both zircote/php-swagger and jms/serializer to annotate my API.  What I'm seeing is that the moment a php-swagger annotation is processed then all jms/serializer annotations are seemingly lost.  Removing all php-swagger annotations fix this.  Reordering the annotations results in anything after the first php-swagger annotation failing instead.  All jms/serializer annotations before the first php-swagger annotation is fine.  I must be doing something wrong as I've seen others use both but I'm unsure what the issue could be.

This is with PHP 7.2 and the latest of zircote/php-swagger and jms/serializer installed via composer.  I tried earlier versions of both packages but the results were the same.

```
use OpenApi\Annotations as OA;
use JMS\Serializer\Annotation\AccessType;
use JMS\Serializer\Annotation\AccessorOrder;
use JMS\Serializer\Annotation\SerializedName;
use JMS\Serializer\Annotation\VirtualProperty;

/**
 * @AccessorOrder("alphabetical")
 * @AccessType("public_method")
 *
 * Class BaseResponse
 *
 * @OA\Schema(
 *     description="Base response",
 *     title="Base response",
 *     schema="BaseResponse"
 * )
 */
class BaseResponse implements IResponse {
        var $errors;

        /**
         * @VirtualProperty
         * @SerializedName("foo")
         *
         * @return string
         */
        public function getFoo() {
                return "foo";
        }
```

> PHP Fatal error: Uncaught Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@ JMS\Serializer\Annotation\VirtualProperty" in method [redacted]\BaseResponse::getFoo() was never imported. Did you maybe forget to add a "use" statement for this annotation? in [redacted]/releases/20200812082918/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationException.php:39

PetrenkoAnton:
@rydan313 Hi

Is there any autocomplete/format plugin for the @OA-annotations for PHPStorm IDE?

DerManoMann:
This should work - so a few questions:
* How do you actually run swagger-php? As part of a bundle or stand-alone?
* What version of doctrine do you use?

It might be the setup of the doctrine autoloader or similar.

You might want to try to add the JMS/Serializer annotations to the swagger-php whitelist:

```
  Analyser::$whitelist = ['JMS\Serializer\\'];
```

rydan313:
[...]

PR #889, opened 2020-12-18
Oo api

DerManoMann:
Adds a OO style way to scan for OpenAPI annotations.

* Adds a new `Generator` class that allows to configure and scan in a non static way
  * Fully BC
  * Only modifies static settings for the duration of the run
  * Allows to override all aspects using fluid setter methods
  * PSR-3 logger support
  * `Generator::scan() accepts` any iterator over source files incl. a Symfony Finder instance
* Migrate to using PSR/LoggerInterface for all logging

To be clear, this is just a proposal and I'd be honestly curious and interested in feedback.

The logger related changes might be controversial (@bfanger and I discussed this before), but I felt this was needed to make the Generator implementation complete.
Test have been updated to allow to test both the default behaviour and injection of a PSR logger instance via an environment variable.
```
export NON_TRACKING_LOGGER=NULL
```

Performance wise this should not have a big impact since ideally there is only a single logger instance that needs to be created upfront.

### DIFF
--- a/tests/OpenApiTestCase.php
+++ b/tests/OpenApiTestCase.php
@@ -217,7 +217,7 @@ class OpenApiTestCase extends TestCase
 
     public function analysisFromFixtures($files): Analysis
     {
-        $analyser = new StaticAnalyser($this->getLogger());
+        $analyser = new StaticAnalyser(null, $this->getLogger());
         $analysis = new Analysis([], null, $this->getLogger());
 
         foreach ((array) $files as $file) {
@@ -229,7 +229,7 @@ class OpenApiTestCase extends TestCase
 
     public function analysisFromCode(string $code, ?Context $context = null)
     {
-        return (new StaticAnalyser($this->getLogger()))
+        return (new StaticAnalyser(null, $this->getLogger()))
             ->fromCode("<?php\n".$code, $context ?: new Context([
                 'logger' => $this->getLogger(),
             ]));
--- a/tests/OpenApiTestCase.php
+++ b/tests/OpenApiTestCase.php
@@ -9,6 +9,7 @@ namespace OpenApi\Tests;
 use DirectoryIterator;
 use Exception;
 use OpenApi\Analyser;
+use OpenApi\Analyser\DocBlockParser;
 use OpenApi\Analysis;
 use OpenApi\Annotations\AbstractAnnotation;
 use OpenApi\Annotations\Info;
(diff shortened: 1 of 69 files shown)


bfanger:
Create a `dont-run-this-script.php` to the fixtures with:

```php
<?php

/**
 * @OA\Delete(path="admin/database", @OA\Response(response="default", description="run drop database script"))
 */
function dropDatabase  {
  throw new Exception("Oh no, I was executed!");
}

dropDatabase();
```

And verify it that it doesn't run when using static analysis.

bfanger:
Not yet fully BC according to travis? ;)

Doqnach:
I really like the concept, where we get rid of those weird static whitelists etc.

DerManoMann:
> 
> 
> Not yet fully BC according to travis? ;)

That's why it is a draft :) I think it would be, but it is hard to catch all the places where a logger has to be passed around.
I think I need to update the tests too to get better tracking of where loggers are not properly passed on yet, but I wanted to get some feedback first before spending more time on that.

I think we should deprecate the static config vars if this PR gets merged to encourage people to use this and then eventually we could remove the static config completely.

DerManoMann:
> 
> 
> I really like the concept, where we get rid of those weird static whitelists etc.

Not weird, just static :)  Customiing or isolating code execution is always hard when static code is in play and a library should try to make things a side effect free as possible.
There are also more types of input supported in `Generator::scan` which hopefully helps adoption.

tests/OpenApiTestCase.php
```diff
@@ -16,45 +15,27 @@
...
-                if (isset($map[$type])) {
-                    $this->fail('Unexpected \OpenApi\Logger::'.$map[$type].'("'.$entry.'")');
-                } else {
-                    $this->fail('Unexpected \OpenApi\Logger->getInstance()->log("'.$entry.'",'.$type.')');
-                }
-            }
-        };
         parent::setUp();
+
+        // reset so processors get new loggers each run
```
DerManoMann, line 34: Realizing this was by far the biggest pain point in this PR! :facepalm:

DerManoMann:
So, I really would like to get this merged at some point soon. One thing outstanding would be to try to run this branch against the current test-suite and confirm that there are is no BC breakage.

I also think this change deserves a new version (at least minor), so that would make it `2.2` I suppose.

On top of that, I would propose to really get rid of all static code (properties/settings) and also the static doctrine `AnnotationRegistry` init. That should allow to run swagger-php side-effect free and avoid any clashes with other libraries, for example: #880. However, that would be a clear BC issue and I'd not want to do that without any other input.

Doqnach:
option would be to first leave the statics in, but with an `@deprecated` warning?

bfanger:
Changing the public api from `$openapi = \OpenApi\scan()`  to `$openapi = (new \OpenApi\Generator())->scan()` is not an improvement.

DerManoMann:
> Changing the public api from `$openapi = \OpenApi\scan()` to `$openapi = (new \OpenApi\Generator())->scan()` is not an improvement.
> 
Well, that was one of the main points of this PR - not having a static function but an OO approach. Its not just about this one line, but the fact that you can configure things on an instance, not static properties.
Not sure what the exact issue is here - more to type? I think for most users this is never going to be an issue as they will either use the CL or a framework bundle/plugin/extension.
Still, I am not oo worried about providing a (less typed) static method...

> I haven't looked at the generator class in detail, but could we make it a facade ([like in laravel](https://laravel.com/docs/8.x/facades)) where you call:
> `\OpenApi\Generator::scan(...)` that way the api looks static, but the implementation is dynamic.
> 
Uhu, I am not a fan of facades and since we most likely will only ever have a single implementation it seems overkill. Again, adding a static scan method to the class should be fine.

> This would allow composer to autoload (functions.php are now always loaded, even when you don't use swagger-php)
>
Yep, that is one reason :)

> We switched to semver so definitly a breaking change and version bump. a little upgrade guide in release notes and a link in the readme.
> A nicer DX would be to throw an Exception then calling \OpenApi\scan with a link to the release notes/migration guide
>
Its a draft and I think I'll leave this PR here for a bit longer for reference but break things down into smaller PRs to actually merge. In particular the logging stuff might need its own PR as rebasing is currently a nightmare :)

PR #895, opened 2021-01-07
Limit `AugmentSchemas` processor to actual `Schema` instances only

### LINKED ISSUE #890: Augment Schemas Processor bug with Items
```php
class AugmentSchemas
{
    public function __invoke(Analysis $analysis)
    {
        $schemas = $analysis->getAnnotationsOfType(Schema::class);
        // Use the class names for @OA\Schema()
            if ($schema->schema === UNDEFINED) {
                if ($schema->_context->is('class')) {
                    $schema->schema = $schema->_context->class;
                } elseif ($schema->_context->is('interface')) {
                    $schema->schema = $schema->_context->interface;
                } elseif ($schema->_context->is('trait')) {
                    $schema->schema = $schema->_context->trait;
                }
            }
        }
```

This part of the AugmentSchemas processor also changes the Items class, as the items class extends Schema and has schema property set to undefined.

This invalidates the Items class as it should not have schema property set and fails the validation with swagger-cli.

DerManoMann:
Set the `strict` flag when calling `getAnnotationsOfType()` to only process actual `Schema` instances.

Fixes #890

### DIFF
--- a/Examples/misc/misc.yaml
+++ b/Examples/misc/misc.yaml
@@ -25,7 +25,6 @@ paths:
                 properties:
                   type: { type: string }
                   color: { type: string }
-                type: object
       responses:
         '200':
           description: Success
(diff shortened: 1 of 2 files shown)


DerManoMann:
Closing for now as it is not clear that the results are correct. Also, the reporter does not need the change after all :(

PR #916, opened 2021-03-16
hash operationId from generated swagger

yagami271:
When i generated the first time a swagger doc. the operationId was indicate tree structure of our class 
like :  `App\Foo\Faa\ClassController::__invoke`

I think we should hide tree structure for a  public doc 

The first and quick idea i found, is to hash the operationId

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /docs/node_modules/*
 .phpunit.result.cache
 /.php_cs.cache
+.idea/
--- a/Examples/using-interfaces/using-interfaces.yaml
+++ b/Examples/using-interfaces/using-interfaces.yaml
@@ -7,7 +7,7 @@ paths:
     get:
       tags:
         - Products
-      operationId: 'UsingInterfaces\ProductController::getProduct'
+      operationId: '330fe5a2e8123ccccf5dac56f09e6ce0'
       parameters:
         -
           name: id
--- a/Examples/using-refs/using-refs.yaml
+++ b/Examples/using-refs/using-refs.yaml
@@ -7,7 +7,7 @@ paths:
     get:
       tags:
         - Products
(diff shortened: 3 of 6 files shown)


DerManoMann:
I like the general idea and reasoning. Not so sure I would want this as he default, but perhaps as an optional feature?

yagami271:
> I like the general idea and reasoning. Not so sure I would want this as he default, but perhaps as an optional feature?

you mean option ? like 
```php
$options = [
    'hashOperationId' => true, (default false)
];

$openapi = OpenApi\scan(__DIR__ . '/app', $options);
```

DerManoMann:
Yes, something like that. Right now there isn't a good place for it, IMO, though. Maybe it is time go get this baby merged :) 
https://github.com/DerManoMann/swagger-php/blob/oo-api/src/Generator.php

yagami271:
> Yes, something like that. Right now there isn't a good place for it, IMO, though. Maybe it is time go get this baby merged :)
> https://github.com/DerManoMann/swagger-php/blob/oo-api/src/Generator.php

You lost me 😂 . i thinks it's will take too much time and to much changes for small change 🤕  to  add optional HashOperationId (as you say " there isn't a good place for it"  ) but the idea is here. hope someone find à sexy solution :)

DerManoMann:
Sorry, I didn't mean to scare you or suggest you merge in thise changes!

I just haven't had time to look deeper into an immediate solution; I'll try to find more time in the next few days. But you are right - it is a good idea one won't go away again. If you are ok we'll leave the PR open for others to see and to remind me to not forget it :)

Either way, thanks for taking the time to help improve swagger-php.

Tobion:
I think it's better to disable this processor instead, see #930

DerManoMann:
Thanks @Tobion

Issue #918, opened 2021-03-24
Nullable refs


raziel057:
Hello,

Using NelmioApiDocBundle I currently try to add a nullable property like that:

```php
   /**
     * @var Status|null
     *
     * @OA\Property(ref=@Model(type=PTC\ApiBundle\TOEntity\Status::class), nullable=true)
     */
    private $status;
```

But when generating the API doc the resulting Json is 

```json
"status":{
    "$ref":"#\/components\/schemas\/Status"
}
```

This is comming from the following code fragment:

```php
// $ref
if (isset($data->ref)) {
    // OAS 3.0 does not allow $ref to have siblings: http://spec.openapis.org/oas/v3.0.3#fixed-fields-18
    $data = (object) ['$ref' => $data->ref];
}
```

Added in the folowing commit:
https://github.com/zircote/swagger-php/commit/113c00a5503b06a143556da596afa594d48c7e6d#diff-96f680a8337541116efb908ea9d65df5fc092a068cce9b6935c638ecc4ee61ad

At this stage we loose the ``nullable`` property from $data. Is it expected? While I generate the REST API Client from .net app, I always need to add manually the nullable true like that, else the property is specified explicitly not nullable:

```json
"status":{
    "$ref":"#\/components\/schemas\/Status",
    "nullable": true
}
```

From the http://spec.openapis.org/oas/v3.0.3#fixed-fields-18 spec it's not very clear how to define nullable refs but as it's taken into account by client generators, I think we should add the property. If good for you I can create a PR with the fix.

DerManoMann:
I've started looking at OpenAPI 3.1.0 and I hope there is a solution in it for this issue to...

DerManoMann:
With attribute feature work wrapping up it seems I should get back to this...

DerManoMann:
This could be a follow up from #1030 although I am not too keen on breaking the spec just because someone else doesn't follow it...

Xerkus:
> I've started looking at OpenAPI 3.1.0 and I hope there is a solution in it for this issue to...

in 3.1.0  nullable is deprecated and `null` is one of the types. For this specific case json schema rules kick in and the result should be something like this:
```
"status":{
    "anyOf": [
      {
       "$ref":"#\/components\/schemas\/Status"
      },
      {
        "type": "null"
      }
    ]
}
```

DerManoMann:
This should be fixed now.

PR #920, opened 2021-03-24
Fix for php < 8.0 (T_ATTRIBUTE doesn't exists)

raziel057:
The constant ``T_ATTRIBUTE`` doesn't exists in 7.x PHP versions.

In composer we can see that the master branch target PHP "php": ">=7.2"


DerManoMann:
This PR contains also changes for issue #918. Could those please be removed? Also, it looks like the build pipeline is broken (well, it should fail...) so I'll be fixing that first...

PR #923, opened 2021-03-25
Ignore invalid utf8 characters

muddydog:
If something with illegal utf8 characters are mixed in your swagger doc. Then the swagger.json or swagger.yaml generated by openapi script will  be empty file. In order to fix this problem, the following changes have been made:
1. Remove emoji; 
2. Use JSON_INVALID_UTF8_IGNORE option to call json_encode function.

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -234,7 +234,7 @@ abstract class AbstractAnnotation implements \JsonSerializable
             $flags = Yaml::DUMP_OBJECT_AS_MAP ^ Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE;
         }
 
-        return Yaml::dump(json_decode($this->toJson(0)), 10, 2, $flags);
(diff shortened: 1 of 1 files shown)


DerManoMann:
I can understand the use of `JSON_INVALID_UTF8_IGNORE` but I wonder if the removal of the emoji is really necessary?

If it leaks into the JSON/Yaml then something is wrong with the underlying code as `UNDEFINED` is not supposed to be persisted. Is there a case where that happens? If so it would be great if you could provide an example of that.

muddydog:
> I can understand the use of `JSON_INVALID_UTF8_IGNORE` but I wonder if the removal of the emoji is really necessary?
> 
> If it leaks into the JSON/Yaml then something is wrong with the underlying code as `UNDEFINED` is not supposed to be persisted. Is there a case where that happens? If so it would be great if you could provide an example of that.

Agree with you, the `UNDEFINED` properties are ignored in JSON/Yaml output. 

When I found this little emoji in the code, it was a pleasant surprise. Remove it from the code, just because of my little cleanliness. So just forget that and keep the easter egg, after all, it does not affect the function.

muddydog:
I have updated my PR

DerManoMann:
Thanks @muddydog

PR #928, opened 2021-04-06
OO API - Take II

DerManoMann:
A stripped down version of PR #889 .

Things missing
* ~documentation~
* some more tests?
* ~doctrine `AnnotationRegistry` loading/unloading~ No use as there is no way to unload other than to clear the registry...


DerManoMann:
A much simpler version of #889. Includes the same features in the `Generator` class but also includes a static `Generator::scan` method that mirrors the current `OpenApi\scan()` function.

Still uses the `Generator` instance as the source of truth for all configurable things :)

Example use:
```
        $openapi = (new Generator())
            ->generate(['src']);

       // or

        $openapi = (new Generator())
            ->generate(Util::finder('src'));

       // or

        $openapi = Generator::scan('src');
```
```
        $openapi = (new Generator())
            ->setAliases(['xx' => '\\My\\Annotations\\'])
            ->generate(['src']);

       // or

         Analyser::whitelist['xx'] = '\\My\\Annotations\\';
        $openapi = Generator::scan('src');
```

DerManoMann:
@bfanger would this be more to your liking?

I still think the overhead of writing a few more chars is really an issue - typical users will probably nerver need to do that all as they either use the command line or some framework bundle/plugin.

Anyway. I hope this PR strikes a good balance between keeping things the same (for now, although deprecated) and allowing to easier manage the current static config.

bfanger:
👍
I think moving some of documentation in "getting started" could be in a migration.md or/and in the release notes.

DerManoMann:
> I think moving some of documentation in "getting started" could be in a migration.md or/and in the release notes.

Agreed. It felt a bit much in both getting started and index but then I couldn't well leave it out.

DerManoMann:
@bfanger (or anyone else, actually) Time for some proof reading? :)

jamesaspence:
As a follow up to my previous comment - I found a way to achieve what I needed by providing a `Finder` of my own to `scan`. Having said that, being able to define include paths without configuring a finder would be a bit nicer of an API :)

DerManoMann:
> As a follow up to my previous comment - I found a way to achieve what I needed by providing a `Finder` of my own to `scan`. Having said that, being able to define include paths without configuring a finder would be a bit nicer of an API :)

Well, I would argue that an API that reuses existing code (ie. `Finder`) is better than another custom implementation... :)

The `Finder` component has documentation, test cases and surely covers more special cases than some random API would implement to read arbitrary files...

bfanger:
Docs looked good 👍 

 `$sources` makes the multiple include paths easier to discover.
The`$directory` parameter for the old scan function also accepts an array of paths, but reading james's comments this demonstrates that this feature was too well hidden ;)

DerManoMann:
KK - so unless there are any other objections I'll be merging early next week and then update gh-pages after that 🎉

jamesaspence:
> > As a follow up to my previous comment - I found a way to achieve what I needed by providing a `Finder` of my own to `scan`. Having said that, being able to define include paths without configuring a finder would be a bit nicer of an API :)
> 
> Well, I would argue that an API that reuses existing code (ie. `Finder`) is better than another custom implementation... :)
> 
> The `Finder` component has documentation, test cases and surely covers more special cases than some random API would implement to read arbitrary files...

Definitely! Don't get me wrong, I don't want to argue to define a new API necessarily, but it'd be nice to have a builder pattern wrapping around these finders or be able to define paths that would translate to finders.

PR #941, opened 2021-05-20
Regression: Fix traits without class-level phpDocs

Yoshi2889:
Release v3.2.0 breaks traits without class-level phpDocs. A workaround is to set an empty phpDoc with just a description. This was not an issue with v3.1.0.

Context::annotations is null in cases where class-level phpDocs don't exist. This causes foreach() to throw a warning which in certain environments causes a full crash.


DerManoMann:
Looking good - thanks. I'll have a bit more of a play tomorrow; if not it looks good to merge.

PR #942, opened 2021-05-26
Operation id improvements

DerManoMann:
Alternative to #916 that allows to configure the behavior of the `OperationId` processor.

Also includes a new method to easily allow updating/replacing a processor.

```
    /**
     * Update/replace an existing processor with a new one.
     *
     * @param callable      $processor the new processor
     * @param null|callable $matcher   Optional matcher callable to identify the processor to replace.
     *                                 If none given, matching is based on the processors class.
     */
    public function updateProcessor(callable $processor, ?callable $matcher = null): Generator
```

```
$generator = (new Generator())
            ->updateProcessor(new OperationId(false));
```

### DIFF
--- a/Examples/using-interfaces/using-interfaces.yaml
+++ b/Examples/using-interfaces/using-interfaces.yaml
@@ -7,7 +7,7 @@ paths:
     get:
       tags:
         - Products
-      operationId: 'GET::UsingInterfaces\ProductController::getProduct'
+      operationId: 6a7963b92da560b8d7f16b0fdba847ef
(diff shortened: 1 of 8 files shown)


Tobion:
> adds the HTTP method to the generated operation id to be unique in case of a controller handling multiple methods

You can also have different paths handled by the same controller, e.g. as aliases or redirects or for BC. So this is still not ensuring unique operation ids as required by the spec.

DerManoMann:
#946

PR #947, opened 2021-06-03
Added example with line-break

winternet-studio:


Examples/misc/misc-api.php
```diff
@@ -16,7 +16,9 @@
 /**
  * @OA\Info(
  *   title="Testing annotations from bugreports",
- *    version="1.0.0"
+ *   version="1.0.0",
+ *   description="NOTE:
+This sentence is on a new line",
```
DerManoMann, line 21: The trailing ',' here is invalid - never quite sure when doctrine ignores it or chokes on it.

DerManoMann (changes requested):
Looking good except for the stray ','. Thanks!

DerManoMann:
Oh, and if you could squash your commits that would be great.

Examples/misc/misc-api.php
winternet-studio: In all my code I made it a habbit to simply add a comma after every single line and I have never had Doctrine complain about it - so that's why I added it :) But I have removed it from here now :)

winternet-studio:
> Oh, and if you could squash your commits that would be great.

Done

DerManoMann:
Thanks @winternet-studio

PR #948, opened 2021-06-10
Added oneOf examples

winternet-studio:


DerManoMann:
Thanks, @winternet-studio

PR #967, opened 2021-07-07
Fix exclusive min/max property type

### LINKED ISSUE #966: exclusiveMaximum/Minimum differnece with JSON Schema Validation
According to http://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.2.3 type of exclusiveMaximum/exclusiveMinimum field must be numeric and represent exclusive maximum value not a boolean. Is there any reason to keep it as a boolean type?

```php
/**
     * See http://json-schema.org/latest/json-schema-validation.html#anchor17.
     *
     * @var bool
     */
    public $exclusiveMaximum = Generator::UNDEFINED;
```

DerManoMann:
PHPdoc documented type needs to be `number` not `bool`.

Fixes #966

### DIFF
--- a/src/Annotations/Schema.php
+++ b/src/Annotations/Schema.php
@@ -116,7 +116,7 @@ class Schema extends AbstractAnnotation
     /**
      * See http://json-schema.org/latest/json-schema-validation.html#anchor17.
      *
-     * @var bool
+     * @var number
      */
     public $exclusiveMaximum = Generator::UNDEFINED;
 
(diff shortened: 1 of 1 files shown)


AleksandrPB:
There is no merge? Why?

DerManoMann:
Ah, are you pulling dev-master via composer?
I might have forgotten to update packagist... Unfortunately auto-update is not configured :/

Composer should now pull the latest master.

PR #979, opened 2021-08-13
+ "excludeProperties" argument is added to "Schema" annotation

philfromfmf:


DerManoMann:
Could you please explain the motivation for this PR? Why not just remove annotations from the properties you do not want?

DerManoMann:
Three more things: 
* For completeness the XMLContent annotation should be updated too
* Updating one of the Examples would be nice
* Why does the old way not work any more? The properties are alread/still there, or not?

PR #980, opened 2021-08-17
Adds examples to nested annotations in JsonContent

jeremy-smith-maco:
Previously we were able to use the `examples={}` syntax for specifying multiple examples for the `JsonContent` tag but now that is not allowed so we have to use `@OA\Examples` instead which is fine but it does not currently render as a dropdown of examples like it used to previously. This change should add back that functionality and allow the dropdown of examples to appear.

### DIFF
--- a/src/Annotations/JsonContent.php
+++ b/src/Annotations/JsonContent.php
@@ -41,5 +41,6 @@ class JsonContent extends Schema
         Property::class => ['properties', 'property'],
         ExternalDocumentation::class => 'externalDocs',
         AdditionalProperties::class => 'additionalProperties',
+        Examples::class => ['examples', 'example'],
     ];
 }


DerManoMann:
A few things: 
* For completeness the XMLContent annotation should be updated too
* Updating one of the Examples would be nice
* Fixing the test ist required 🙂
* Why does the old way not work any more? The properties are alread/still there, or not?

jeremy-smith-maco:
Yep, I will update those things. 

> Why does the old way not work any more? The properties are alread/still there, or not?

It doesn't allow you to use `examples={ ... }`. It throws an error telling you to use `@OA\Examples` instead.

DerManoMann:
> > Why does the old way not work any more? The properties are alread/still there, or not?
> 
> It doesn't allow you to use `examples={ ... }`. It throws an error telling you to use `@OA\Examples` instead.

Oh, ok. It just sounded like it used to work before...

jeremy-smith-maco:
> > > Why does the old way not work any more? The properties are alread/still there, or not?
> > 
> > 
> > It doesn't allow you to use `examples={ ... }`. It throws an error telling you to use `@OA\Examples` instead.
> 
> Oh, ok. It just sounded like it used to work before...

It did used to work before. We updated the library from `3.1.0` to `3.2.3` and that seems to have caused it start throwing an exception about it. It is throwing:

```
@OA\MediaType(mediaType="application/json")->examples is an object literal, use nested @OA\Examples() annotation(s) in ...
```

DerManoMann:
Looking good to me. I did some validation on the updated specs and at least `spectral` doesn't like it.
I think its wrong (https://github.com/stoplightio/spectral/issues/1780) but I'd like to give that ticket a couple days at least before merging...

PR #982, opened 2021-08-26
PHP 8.1 attributes

DerManoMann:
Sneak preview on attributes (requires PHP 8.1).

In fact, this is now the branch for swagger-php v4!

Feedback welcome!

So far:
* bumps major version to 4.x
* moved/renamed `TokenAnalyser`and `Analyser`
* introduced a new `ReflectionAnalyser` and `AnnotationFactoryInterface`
* introduced new `AnalyserInterface` implemented by both `TokenAnalyser` and `ReflectionAnalyser`
* new annotation factories added: `DocBlockAnnotationFactory` and `AttributeAnnotationFactory`; `ReflectionAnalyser` can use either (or possibly both at the same time...)
* updated all examples to pass with both token analyser and reflection analyser (using `DocBlockAnnotationFactory`)
  - this required adding classes to lots of stand-alone docblocks....
* Updated some annotation classes to also work as `\Attribute`
* Attribute based fixture in`tests/Fixtured/Apis/Attributes` - this is on parity with the corresponding DocBlock example
* Allow to mix attributes and annotations
* Add PHP 8.1 to build matrix (latest deps only)

TODO:
* add all supported properties as named ctor parameters to PHP 8.1 constructors
* add more attribute based tests
* things I've missed to far :)


DerManoMann:
Progress: The openapi-spec example now passed with all analysers!
https://github.com/DerManoMann/swagger-php/tree/php81-attributes/Examples/openapi-spec

Unfortunately the nested attributes have to be wrapped into a single line to be BC.

Tobion:
From what I see, you want to support both attributes and annotations? Wouldn't it be easier and better to just support attributes in the next major version? 
To me, either you try to make attributes and annotations interchanable without bc break. Then there is no reason for a new major version. Or you create a new major version and then there is not need to support legacy annotations anymore.

DerManoMann:
> To me, either you try to make attributes and annotations interchanable without bc break. Then there is no reason for a new major version. Or you create a new major version and then there is not need to support legacy annotations anymore.

I think we do have a different understanding of semantic versioning. To me the version bump is justified by the big refactor of existing classes. I doubt that any downstream project using swaggger-php could update to this new code without some  work.

Having a new major version doesn't mean we do have to abandon all annotation support. Considering that apparently people still use version 2 ;) I would expect lots of projects will have other priorities than updating annotations.
It also means regular users should be able to upgrade without issues and we do not have to maintain three main versions at the same time.

The way I see it anyone just using swagger-php via CL should be able to upgrade without pressure of migrating to attributes (which would require updating the project to PHP 8.1, no less).
Anyone using something like https://github.com/nelmio/NelmioApiDocBundle will be stuck with whatever that project does anyway and I do not expect those to drop annotation support quickly either.

Finally, being able to run both side by side (or mixed, which I think is possible) helps immensly with testing. There are a lot of tests that would not work using the new `ReflectionAnalyser` and migrating those will take some time. Right now I do not expect that to be fully done before this new version is released.

Beside all that I'd be still interested in actual feedback (testing, things missing, etc...)

DerManoMann:
@SlvrEagle23 I've added some more type hints to the 8.1 named attributes - could you guy have a look and see if that helped things?

Also, those are only enabled with PHP 8.1, not 8.0 as some attributes really require nesting...

Anything else, just let me know; I know there are some named properties in the attribute constructors still missing - if you find any please let me know ( or any other issues, for that matter)

SlvrEagle23:
So, I keep a pretty close eye on the PHP internals and I haven't actually seen anything at all about allowing nested attributes in PHP 8.1, and I looked through the approved RFCs and found nothing of the sort either.

It's entirely possible I'm missing something, though. I'd certainly love to see it implemented in 8.1 but I'm not seeing any evidence at the moment that it's coming down the line.

DerManoMann:
Ah, my mistake for not bein precise. Its not really nested attributes but you can use `new` in the attribute parameters.

Depending on how details you annotate nesting is the only sane way; some of the tests and examples do it, for example https://github.com/DerManoMann/swagger-php/blob/php81-attributes/tests/Fixtures/Apis/Attributes/basic.php#L44 and https://github.com/DerManoMann/swagger-php/blob/php81-attributes/Examples/openapi-spec-attributes/UsersController.php#L28

SlvrEagle23:
Aha, indeed it appears `new` is now allowed in attributes as of 8.1, basically allowing nesting that way. A very welcome addition!

itsjavi:
@DerManoMann due to the duplication of classes, PHPStorm 2021.2.3 is not able to find the attribute ones
Screenshot 2021-12-11 at 01 26 40

If I comment out the Annotation class and leave the Attribute one, all it's fine:

Screenshot 2021-12-11 at 01 27 12


Wouldn't it be better to define the attributes in a new namespace?

DerManoMann:
Honestly, I do not know why that would be. I have a project using 4.0.4 attributes and that is fine.

I do get a message about multiple implementations but that is it. Again, the configured language level might make a difference.

I do agree that a separate namespace would be nice. In fact, that is how I started. Unfortunately, some fo the internals do check for explicit class names to verify allowed parent/child relationships and that would have been a major pain.

itsjavi:
@DerManoMann I have the target language as php 8.1. It might be just a phpstorm issue, but anyway having duplicated class names is always tricky because it may get other analysis tools "confused" as well.

thanks for the effort in this PR anyway

DerManoMann:
Thanks @itsjavi; also apologies for avoiding more work :)

Happy to say that in light of #1021 I've started looking into a potential split... Turns out some very early iteration had some code to deal with different namespaces for attributes/annotations already - I just thought that never made it into the real PR but it seems I was wrong :)

So, chances are the change might not be too bad after all.

DerManoMann:
> Why not 8.0?

Except for simple attributes there is some nesting required with using `new ...` and that is only available in 8.1

PR #984, opened 2021-09-15
fix php 8.1 deprecations in 2.x

Tobion:
#981 for 2.x


rbatukaev:
@DerManoMann Would you please create a new tag for `2.x`?

PR #998, opened 2021-11-11
handle nullable typed arrays

dbu:
i noticed that we had some fields in our openapi doc that had no type. with some digging, i realized that it was the case for all things that are a typed array that is also nullable. for example `Customer[]|null`.

as a workaround, it works correctly when flipping the order to `null|Customer[]`. but that seems unexpected.

i added a test to prove the problem, and a proposal how to fix it.

### DIFF
--- a/src/Processors/AugmentProperties.php
+++ b/src/Processors/AugmentProperties.php
@@ -83,9 +83,12 @@ class AugmentProperties
                 }
             } elseif (preg_match('/@var\s+(?<type>[^\s]+)([ \t])?(?<description>.+)?$/im', $comment, $varMatches)) {
                 if ($property->type === Generator::UNDEFINED) {
-                    preg_match('/^([^\[]+)(.*$)/', trim($varMatches['type']), $typeMatches);
-                    $isNullable = $this->isNullable($typeMatches[1]);
-                    $type = $this->stripNull($typeMatches[1]);
+                    $allTypes = trim($varMatches['type']);
+                    $isNullable = $this->isNullable($allTypes);
+                    $allTypes = $this->stripNull($allTypes);
+                    preg_match('/^([^\[]+)(.*$)/', trim($allTypes), $typeMatches);
+                    $type = $typeMatches[1];
+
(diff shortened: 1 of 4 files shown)


src/Processors/AugmentProperties.php
```diff
@@ -83,9 +83,12 @@ public function __invoke(Analysis $analysis)
                 }
             } elseif (preg_match('/@var\s+(?<type>[^\s]+)([ \t])?(?<description>.+)?$/im', $comment, $varMatches)) {
                 if ($property->type === Generator::UNDEFINED) {
-                    preg_match('/^([^\[]+)(.*$)/', trim($varMatches['type']), $typeMatches);
-                    $isNullable = $this->isNullable($typeMatches[1]);
-                    $type = $this->stripNull($typeMatches[1]);
+                    $allTypes = trim($varMatches['type']);
+                    $isNullable = $this->isNullable($allTypes);
+                    $allTypes = $this->stripNull($allTypes);
```
dbu, line 88: if we first check nullable and then parse for array notation, it seems to work.

tests/Processors/AugmentSchemasTest.php
```diff
@@ -26,7 +26,7 @@ public function testAugmentSchemas()
         $this->assertSame(Generator::UNDEFINED, $customer->properties, 'Sanity check. @OA\Property\'s not yet merged ');
         $analysis->process(new AugmentSchemas());
         $this->assertSame('Customer', $customer->schema, '@OA\Schema()->schema based on classname');
-        $this->assertCount(9, $customer->properties, '@OA\Property()s are merged into the @OA\Schema of the class');
+        $this->assertCount(10, $customer->properties, '@OA\Property()s are merged into the @OA\Schema of the class');
```
dbu, line 29: i added a new property with the nullable and typed array annotation.

DerManoMann:
Thanks @dbu

dbu:
np. thanks for this library, it is useful for us!

dbu:
could you do another patch release with this fix please?

PR #999, opened 2021-11-12
Fix arguments exception, resolves #997

### LINKED ISSUE #997: Argument 2 passed to OpenApi\Analysis::addAnnotations() must be an instance of OpenApi\Context
Since upgrading to 3.3.0 yesterday, I get the following error when creating the documentation:

`In Analysis.php line 128: Argument 2 passed to OpenApi\Analysis::addAnnotations() must be an instance of OpenApi\Context, null given, called in .../vendor/zircote/swagger-php/src/Analysis.php on line 85`

Works fine after downgrading again:
`Downgrading zircote/swagger-php (3.3.0 => 3.2.3)`

antonvolokha:
When I updated lib to version 3.3.0 I saw error like this
```
{"message":"OpenApi\\Analysis::addAnnotations(): Argument #2 ($context) must be of type OpenApi\\Context, null given, called in \/app\/vendor\/zircote\/swagger-php\/src\/Analysis.php on line 82"}
```
It happened when you use this lib with [nelmio/api-doc-bundle](https://github.com/nelmio/NelmioApiDocBundle) for symfony
because they use empty constructor
```
$analysis = new Analysis();
$analysis->addAnnotation($this->openApi, null);
```

so I added nullable support and everything works fine  

resolves #997

### DIFF
--- a/src/Analysis.php
+++ b/src/Analysis.php
@@ -85,7 +85,7 @@ class Analysis
         $this->addAnnotations($annotations, $context);
     }
 
-    public function addAnnotation($annotation, Context $context): void
+    public function addAnnotation($annotation, ?Context $context): void
     {
         if ($this->annotations->contains($annotation)) {
             return;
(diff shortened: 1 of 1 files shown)


DerManoMann:
Well, this might be the pragmatic solution for the moment although ideally this should rather be fixed in the nelmio bunde. As mentioned before, all nelmio tests do pass for me...

While this fixes the immediate issue it will break things in swagger-php in case of processing errors as then a context is expected to be there in order to access the logger...

PR #1009, opened 2021-12-04
Fix autoload issue

leo108:


DerManoMann:
Could you please elaborate a bin on what exact problem you are having using swagger-php?

While it is true that the annotation classes are not perfect due to the split between PHP 8.1 and other versions each file also contains a class with a matching name. So, there is also a class named `OpenApi` in the file.

The current setup was a deliberate choice on my side, so there would have to be at least some justification to change it.

DerManoMann:
Superseded by #1024

Issue #1016, opened 2021-12-07
ErrorException: Uninitialized string offset 1


fuale:
at 4.0.1
```
  at vendor/zircote/swagger-php/src/Analysers/TokenScanner.php:128
    124▕                             // no function body
    125▕                             $this->skipTo($tokens, ';');
    126▕                         }
    127▕ 
  ➜ 128▕                         $units[$currentName]['methods'][] = $token[1];
    129▕                     }
    130▕                     break;
    131▕                 case T_VARIABLE:
    132▕                     if (1 == count($stack) && $currentName) {
```

Place where i found bug using debugger

```
$worklogs->issues = array_map(static function ($issue) use ($client) {
    if ($issue->fields->worklog->total > $issue->fields->worklog->maxResults) {
        $issue->fields->worklog = JiraQuery::getWorklogs($client, $issue->key);
    }

    return $issue;
}, $worklogs->issues);

return collect($worklogs->issues);
```
I Do something wrong?

fuale:
I think it's a problem with nested `static function` statement

fuale:
additional debug info
```
$lastToken = {array} [3]
 0 = {int} 321
 1 = "static"
 2 = {int} 64
$namespace = "App\Services\Jira"
$stack = {array} [1]
 0 = "{"
$token = "("
```

fuale:
Yes, I commented out this piece of code, and now it's working

DerManoMann:
Sigh, thanks. Another test case, another release :/

DerManoMann:
Hmm, there must be something else. I've added this methot to one of the test fixtures and that works fine.
```
    public function whatever($client)
    {
        $worklogs = (object)[];
        $worklogs->issues = array_map(static function ($issue) use ($client) {
            if ($issue->fields->worklog->total > $issue->fields->worklog->maxResults) {
                $issue->fields->worklog = 'foo';
            }

            return $issue;
        }, $worklogs->issues);

        return collect($worklogs->issues);
    }

```

@fuale Would you be able to extract a code snippet into a separate file or something that does break the code and share that? There must be some more context that is missing to make it go wrong.

DerManoMann:
You can see the code I used for a fixture here: 

https://github.com/DerManoMann/swagger-php/blob/more-nested-functions/tests/Fixtures/PHP/AnonymousFunctions.php#L56

fuale:
oh, my fault
```
public function whatever($client)
{
[...]

PR #1017, opened 2021-12-07
Add more function variations to fixture

DerManoMann:


fuale (commented):
I test `static fn() =>` statement too, everything ok #1016

PR #1029, opened 2021-12-17
deprecation notice addressed

Eloar:
Small change to suppress deprecation notice for library since Symfony5 calls for action for that. As library is compatible with PHP versions before 8.0 only option to maintain that is to add `@return mixed` doc annotation instead return type hint.


DerManoMann:
Fixed via #1033 - was easier then fixing the code-style issue... Thanks anyway!

Issue #885, opened 2020-12-14
OpenAPI 3.1 support


DerManoMann:
Reminder/placeholder to collect features we want/need to implement...

psk88:
Add support the style and explode attributes for multipart/form-data that can be used to control array serialization.

tm1000:
Callback:
- https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#callbackObject
Webhooks:
- https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#oasWebhooks

ekisu:
It would be nice to have support for `summary`/`description` attributes on `$refs`
https://github.com/OAI/OpenAPI-Specification/blob/3.1.0/versions/3.1.0.md#reference-object

DerManoMann:
Not really - I spend a lot of time on attributes lately. However, the dust has settled a bit on that so 3.1 might be on the cards next.

xYundy:
SwaggerPHP v4 was released. Any ETA for at least for `$refs` summary/description override? What's next on roadmap?

DerManoMann:
If things were just that easy :)

https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

I guess we'll have to widen some type hints...

DerManoMann:
Closing as mostly done. It will be easier to create new issues for individual features if needed.

PR #1030, opened 2021-12-17
feat: openapi: 3.1.0  summary and description fields for refs

bfanger:
- Changed default version to `openapi: 3.1.0`
- When specified, output the **summary** and/or **description** fields alongside the `ref`

Related to  https://github.com/zircote/swagger-php/issues/885


DerManoMann:
@bfanger I'll fix the conflicts and review today/tomorrow.

DerManoMann:
Looking good to me - will be fun merging into PR #1024 but I guess that is my problem :)

Speaking of problems - `spectral` raises a couple new, 3.1.0 related issues with this PR - I guess we can tackle those in follow up PRs...

DerManoMann:
Ah, forgot to mention that I'd like to merge this and #1024 together and then create 4.1.0 after that unless there is reason that the  spec version change to 3.1.0 should be a major version bump.

DerManoMann (approved):
I suppose it will make a few people happy :)

src/Annotations/AbstractAnnotation.php
```diff
@@ -345,8 +345,17 @@ public function jsonSerialize()
...
         // $ref
         if (isset($data->ref)) {
-            // OAS 3.0 does not allow $ref to have siblings: http://spec.openapis.org/oas/v3.0.3#fixed-fields-18
-            $data = (object) ['$ref' => $data->ref];
+            // Only specific https://github.com/OAI/OpenAPI-Specification/blob/3.1.0/versions/3.1.0.md#reference-object
+            $ref = ['$ref' => $data->ref];
+            $defaultValues = get_class_vars(get_class($this));
+            foreach (['summary', 'description'] as $prop) {
+                if (property_exists($this, $prop)) {
+                    if ($this->$prop !== $defaultValues[$prop]) {
```
DerManoMann, line 353: Just a question: Is there any value in looking up default values this convoluted way? After all we know that `Generator::UNDEFINED`is the default value...

PR #1050, opened 2022-01-04
Add enum support

### LINKED ISSUE #1043: Enums
Now that php has native enums Ian there a plan to add in a native enum attribute?

OpenAPI has enum support in version 3

DerManoMann:
Fixes #1043


ymilin:
Hi,

I downloaded and tried this PR, the case we discussed in the issue works really well 👍 :

```php
#[Schema]
enum Status
{
    case DRAFT;
    case PUBLISHED;
    case ARCHIVED;
}
```
Successfully generates:
```yaml
components:
  schemas:
    Status:
      type: string
      enum:
        - DRAFT
        - PUBLISHED
        - ARCHIVED
```

I then tried to use it in a Schema like this:
```php
#[Schema]
class BlogPost {
    public function __construct(
        #[Property] public int $id,
        #[Property] public string $content,
        #[Property] public Status $status,
    ) {}
}
```
which generates:
```yaml
components:
  schemas:
    BlogPost:
      properties:
        id:
          type: integer
        content:
          type: string
        status: {  }         # <-- should be $ref: '#/components/schemas/Status'
      type: object
    Status:
      type: string
      enum:
        - DRAFT
        - PUBLISHED
        - ARCHIVED
```
Here `$status` in `BlogPost` should be inferred as a reference to #/components/schemas/Status

Manually setting `ref` works:
```php
class BlogPost {
    public function __construct(
        #[Property] public int $id,
        #[Property] public string $content,
        #[Property(ref: "#/components/schemas/Status")] public Status $status, # <-- works
    ) {}
}
```

In a similar manner, I tried doing this:
```php
enum Fizz {
    case FOO;
    case BAR;
    case BAZ;
}

#[Schema]
class Buzz {
    public function __construct(
        #[Property] public int $id,
        #[Property] public Fizz $fizz,
    ) {}
}
```

Which Generates:
```yaml
components:
  schemas:
    Buzz:
      properties:
        id:
          type: integer
        fizz: {  }
      type: object
```
Should Generate:
```yaml
components:
  schemas:
    Buzz:
      properties:
        id:
          type: integer
        fizz:
          type:string
          enum:
            - FOO
            - BAR
            - BAZ
      type: object
```

PR #1052, opened 2022-01-05
Revert default OpenAPI version to 3.0.0

GuilhemN:
Hello,

We have two issues opened on NelmioApiDocBundle due to the bump to OpenAPI 3.1.0: https://github.com/nelmio/NelmioApiDocBundle/issues/1943 (PHPStorm leaks memory with 3.1.0), and https://github.com/nelmio/NelmioApiDocBundle/issues/1940 (Swagger-UI does not support 3.1.0 yet).

I believe the bump was done too early as the ecosystem tools are not ready to support 3.1.0 yet, and in my opinion it would be better to stick to 3.0.0 for the time being.

I opened this PR to share my thoughts, and in my opinion only the default version should be reverted, and not the support of 3.1 specific features as they could be used opt-in. But I would also understand if you don't want to revert this as the bump would likely be more difficult to enforce later for BC reasons, if that's the case we would default back to 3.0.0 in NelmioApiDocBundle until `Swagger-UI` adds support for 3.1.0.

### DIFF
--- a/Examples/example-object/example-object.yaml
+++ b/Examples/example-object/example-object.yaml
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
 info:
   title: 'Example for response examples value'
   version: '1.0'
--- a/Examples/misc/misc.yaml
+++ b/Examples/misc/misc.yaml
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
 info:
   title: 'Testing annotations from bugreports'
   description: "NOTE:\nThis sentence is on a new line"
--- a/Examples/nesting/nesting.yaml
+++ b/Examples/nesting/nesting.yaml
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
 info:
   title: 'Nested schemas'
   description: 'An entity controller class.'
--- a/Examples/openapi-spec-attributes/openapi-spec-attributes.yaml
+++ b/Examples/openapi-spec-attributes/openapi-spec-attributes.yaml
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
 info:
   title: 'Link Example'
   version: 1.0.0
--- a/Examples/openapi-spec/openapi-spec.yaml
+++ b/Examples/openapi-spec/openapi-spec.yaml
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
 info:
   title: 'Link Example'
   version: 1.0.0
--- a/Examples/petstore-3.0/petstore-3.0.yaml
+++ b/Examples/petstore-3.0/petstore-3.0.yaml
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
 info:
   title: 'Swagger Petstore'
   description: "This is a sample Petstore server.  You can find\nout more about Swagger at\n[http://swagger.io](http://swagger.io) or on\n[irc.freenode.net, #swagger](http://swagger.io/irc/)."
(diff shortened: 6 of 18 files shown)


DerManoMann:
Good question. I have to admin I didn't think too much about this (which I perhaps should have).

Opt-in was what I had originally planned  - that is why I never really got tracking on the issue because there were other things do be done first.
I would think that adding a switch to allow to pick the OpenAPI version should be fine in terms of BC - let me have a think...

DerManoMann:
How about #1053 ?

GuilhemN:
#1053 still means that people upgrading to SwaggerPHP 4.1.0 will suddenly have a documentation based on OpenApi 3.1 and it will no longer work in Swagger-UI at least without manual intervention.

For me this is a BC break. But I'd say the inner question is whether this is acceptable, v4 is recent so the user base impacted is allegedly relatively small (25x less daily installs than v3 according to packagist: https://packagist.org/packages/zircote/swagger-php/stats#major/all). Also there's possibly a non negligible fraction coming from NelmioApiDocBundle if they did not put a direct dependency on `zircote/swagger-php` and so were automatically upgraded to v4 and we can fix the break in NelmioApiDocBundle itself by setting the OpenApi version to 3.0 manually.

#1053 doesn't really fix the issue in my opinion though it has the merit of expliciting the support of OpenApi 3.0 and that we would not be "hacking around" by setting the api version in NelmioApiDocBundle if you decide not to revert.

So just to conclude I personally prefer avoiding BC breaks as much as possible and would be in favor of reverting but the youth of v4 and the need to make the lib evolve is a perfectly valid argument in favor of not reverting so I'd understand should you prefer to go this way :-)

DerManoMann:
I am totally with you on BC breaks. Considering that 3.0.0 to 3.1.0 is a minor step I figured this would be acceptable.
Sorry for having let this slip.

The reasoning about why #1053 is not sufficient escapes me a bit, though. Allowing you to set the version would have the same effect as changing the default version string in swagger-php. In both cases the spec would be 3.1.0, or not?

As I've stated in the PR there is room for perhaps making the 3.1.0 changes conditional to the set version, but I figured getting to the same point as your PR in a way that leaves this option would be a good start and get the ball rolling.

DerManoMann:
An alternative could be to make 3.0.0 the default again and allow to opt-in to 3.1.0.

Either way the 3.1.0 changes in the spec would still be there without additional changes.

DerManoMann:
In fact, making things conditionally wasn't too bad, so I've added that to the PR. To be honest, using it like this was one of the main motivations for enforcing the context everywhere in the first place...

GuilhemN:
> I am totally with you on BC breaks. Considering that 3.0.0 to 3.1.0 is a minor step I figured this would be acceptable.
Sorry for having let this slip.

No problem, I would certainly I've done the same :) it's actually quite unexpected that OpenApi own libraries are not ready for 3.1 considering it only contains minor changes.

> The reasoning about why #1053 is not sufficient escapes me a bit, though. Allowing you to set the version would have the same effect as changing the default version string in swagger-php. In both cases the spec would be 3.1.0, or not?

My point is really only about this line: https://github.com/zircote/swagger-php/pull/1053/files#diff-02d5bc58fb2a8f225882e7316cddedeb14fc9d8a589df6624a08bfceebca23e0R24, all the rest is good for me in #1053. Indeed, if the default version remains 3.1, if an user did not set the OpenAPI version manually when using SwaggerPHP 4.0, it will be bumped automatically to 3.1 when upgrading SwaggerPHP 4.0 -> 4.1 which can break its usage if imported in Swagger-UI notably. While if the default version was changed back to 3.0 the upgrade would be completely seamless since the documentation produced would remain the same.

> An alternative could be to make 3.0.0 the default again and allow to opt-in to 3.1.0.
>
> Either way the 3.1.0 changes in the spec would still be there without additional changes.

This is the option I prefer :) in any case I'm just statting my point of view and what I would do but I will be fine with any decision you take and will adapt.

matheusbloise:
can y merge it please?

DerManoMann:
Addresses in #1053 1053

PR #1053, opened 2022-01-05
Allow to set the OpenAPI version

DerManoMann:
Allows to toggle the openapi version with only '3.1.0' and '3.0.0' being allowed.

Right now this will not affect the generated spec, but that could be added later if required without affecting the `Generator`


GuilhemN (approved):
This looks great to me 👍🏻 
Thanks a lot for your reactivity, this was actually much more work than I expected!

PR #1054, opened 2022-01-07
Make Response::description optional when $ref is set

### LINKED ISSUE #1051: `Response` shouldn't require `description` when `$ref` is provided.
Per the documentation on the OpenAPI spec: https://swagger.io/docs/specification/describing-responses/

Specifically this example:

```yaml
paths:
  /users:
    get:
      summary: Gets a list of users.
      response:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ArrayOfUsers'
        '401':
          $ref: '#/components/responses/Unauthorized'   # <-----
  /users/{id}:
    get:
      summary: Gets a user by ID.
      response:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/User'
        '401':
          $ref: '#/components/responses/Unauthorized'   # <-----
        '404':
          $ref: '#/components/responses/NotFound'       # <-----
# Descriptions of common components
components:
  responses:
    NotFound:
      description: The specified resource was not found
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/Error'
    Unauthorized:
      description: Unauthorized
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/Error'
  schemas:
    # Schema for error response body
    Error:
      type: object
      properties:
        code:
          type: string
        message:
          type: string
      required:
        - code
        - message
```

If you provide a `$ref`, it seems like it should just pull everything from the response it's referencing, but the validator in the code right now still requires a "Description" field, even though that's redundant to the one defined in the ref.

DerManoMann:
Fixes #1051

### DIFF
--- a/Examples/misc/misc-api.php
+++ b/Examples/misc/misc-api.php
@@ -47,7 +47,7 @@ class OpenApiSpec
  *      @OA\Property(property="color", type="string"),
  *   )),
  *   security={{ "bearerAuth":{} }},
- *   @OA\Response(response="200",ref="#/components/responses/200", description="Success")
+ *   @OA\Response(response="200", ref="#/components/responses/200")
  * )
  */
 class Endpoint
@@ -57,7 +57,7 @@ class Endpoint
 /**
  * @OA\Response(
  *     response=200,
- *     description="",
+ *     description="Success",
  *     @OA\MediaType(
  *          mediaType="application/json",
  *          @OA\Schema(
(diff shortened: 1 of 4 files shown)


SlvrEagle23:
@DerManoMann Looking at this, the PHP 8.1 attribute version of `Response` now asks for a `ref` but then does nothing with it. This means that if it's all you have here, it won't ever be set and won't ever validate. Probably a bug that needs fixing (and testing against).

PR #1055, opened 2022-01-09
Fixed RequestBody docs

pavel-vasiluk:
- Fixed RequestBody annotation docs


DerManoMann:
Thanks @pavel-vasiluk

PR #1057, opened 2022-01-10
`Response`: Pass ref to parent constructor.

SlvrEagle23:
This fixes the bug where the `ref` attribute is never set on the `Response` object.


DerManoMann:
Thanks @SlvrEagle23

PR #1060, opened 2022-01-13
Improve PHP 8.1 enum handling

### LINKED ISSUE #1056: Support Enum as type hints
Follow up from #1050

I tried to use it in a Schema like this:
```
#[Schema]
class BlogPost {
    public function __construct(
        #[Property] public int $id,
        #[Property] public string $content,
        #[Property] public Status $status,
    ) {}
}
```
which generates:
```
components:
  schemas:
    BlogPost:
      properties:
        id:
          type: integer
        content:
          type: string
        status: {  }         # <-- should be $ref: '#/components/schemas/Status'
      type: object
    Status:
      type: string
      enum:
        - DRAFT
        - PUBLISHED
        - ARCHIVED
```
Here $status in BlogPost should be inferred as a reference to #/components/schemas/Status

Manually setting ref works:
```
class BlogPost {
    public function __construct(
        #[Property] public int $id,
        #[Property] public string $content,
        #[Property(ref: "#/components/schemas/Status")] public Status $status, # <-- works
    ) {}
}
```
In a similar manner, I tried doing this:
```
enum Fizz {
    case FOO;
    case BAR;
    case BAZ;
}

#[Schema]
class Buzz {
    public function __construct(
        #[Property] public int $id,
        #[Property] public Fizz $fizz,
    ) {}
}
```
Which Generates:
```
components:
  schemas:
    Buzz:
      properties:
        id:
          type: integer
        fizz: {  }
      type: object
```
Should Generate:
```
components:
  schemas:
    Buzz:
      properties:
        id:
          type: integer
        fizz:
          type:string
          enum:
            - FOO
            - BAR
            - BAZ
      type: object
```

DerManoMann:
Fixes #1056 

* Expand complex type hints (class/interface/enum) to `$ref` if a corresponding schema exists
* Allow `Property` attribute on method arguments; this allows to annotate promoted properties in models

Allows something like:
```php
    // ...

    items: new OAT\Items(type: Product::class))

    // ...

    #[OA\Get(path: 'foo'])
    public function getFoo(
        #[OA\PathParameter()] string $username,
        #[OA\PathParameter()] State $state
    )
    {
    }
```
instead of
```php
    // ...

    items: new OAT\Items(ref: '#/components/schemas/Product'))

    // ...

    #[OA\Get(path: 'foo'])
    public function getFoo(
        #[OA\PathParameter()] string $username,
        #[OA\PathParameter(ref: '#/components/schemas/State')] State $state
    )
    {
    }
```

Works for classes, interfaces and enums.


Tobion:
The example in the description does not make sense to me. It's using PathParameter but the configured path does not contain any parameters. Is this just a bad example or am I missing something?

DerManoMann:
Bad example. I guess there could be some validation but there is no standard way of how vars in the path are marked (although something like`{id}` seems pretty common

PR #1076, opened 2022-01-19
Change typehint of Operation tags to string

stefanbraspenning:
Tags should be an array of strings, not Tag's. This would be a bc-break and it also breaks the openapi generation


DerManoMann:
Thanks @stefanbraspenning

PR #1079, opened 2022-01-20
Add `minimum`/`maximum` to Schema

### LINKED ISSUE #1078: Missing `minimum`/`maximum` in Schema when using Attributes
It may be me that is missing something, but I can't find `minimum`/`maximum` on Schema anymore after trying to migrate to Attribute.

This is what I had using v3 with Annotations:
```php
/**
 *     @OA\Parameter(
 *         name="per_page",
 *         in="query",
 *         description="Number of records to include per page.",
 *         required=false,
 *         example=100,
 *         @OA\Schema(
 *             type="integer",
 *             default=100,
 *             minimum=1,
 *             maximum=1000
 *         )
 *     )
 */
```

And this is what I try to use with Attributes in v4.2.3:
```php
#[Parameter(
    name: 'per_page',
    description: 'Number of records to include per page.',
    in: 'query',
    required: false,
    schema: new Schema(type: 'integer', default: 100, minimum: 1, maximum: 1000, example: 100),
)]
```

Everything else in that Parameter and Schema seems to work except `minimum` and `maximum`.

I get the following error: `Unknown named parameter $minimum`

Please accept my apology if it's just me that has missed something. That they should be nested or are located somewhere else.

hmazter:
Add minimum and maximum properties to Schema Attribute

Fixes #1078


hmazter:
Fixed by #1080

DerManoMann:
Closing in favour of #1080. Sorry @hmazter, normally I would have happily accepted your PR but in this case there are more properties missing and also for a few more attributes...

hmazter:
No problem! 👍

PR #1085, opened 2022-01-28
X-Logo

BlackRhinoMedia:
Attempting to add support for X-Logo.

I have tried to mimic the License/Contact approach to add support for X-Logo within the info object so in turn I can add support in l5-swagger so it can generate the logo json correctly.

Any feedback - I am sure someone with more knowledge than me could fix it in no time if I have not done it correctly.

### DIFF
--- a/src/Annotations/Info.php
+++ b/src/Annotations/Info.php
@@ -54,6 +54,13 @@ class Info extends AbstractAnnotation
     public $license = Generator::UNDEFINED;
 
     /**
+     * The logo information for the exposed API.
+     *
+     * @var License
+     */
+    public $xlogo = Generator::UNDEFINED;
+
+    /**
      * The version of the OpenAPI document (which is distinct from the OpenAPI Specification version or the API implementation version).
      *
      * @var string
@@ -81,6 +88,7 @@ class Info extends AbstractAnnotation
     public static $_nested = [
         Contact::class => 'contact',
         License::class => 'license',
+        XLogo::class => 'xlogo',
         Attachable::class => ['attachables'],
     ];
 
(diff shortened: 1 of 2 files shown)


DerManoMann:
I don't think that a logo is part of the OpenAPI spec so probably this is outside of the scope of this library.

There is a standard way of adding vendor specific data to the OpenAPI document and that is using the `x` vendor extension.

Each annotation and attribute accepts a parameter `x` which may contain a map so you could do something like this:

```php
/**
 * @OA\Info(
 *   title="Test",
 *   version="1.0.0",
 *   x={"logo-url"="https://localhost/lgo.png"}
 * )
 */
```

This will give you
```yaml
info:
  title: Test
  version: 1.0.0
  x-logo-url: 'https://localhost/lgo.png'
```

DerManoMann:
This would also work

```php
/**
 * @OA\Info(
 *   x={"logo"={"url"="https://localhost/lgo.png"}}
 * )
 */
```

and give you
```yaml
info:
  x-logo:
    url: 'https://localhost/lgo.png'
```

BlackRhinoMedia:
@DerManoMann 

OK thank you for responding so quickly. I received an answer on [Stack Overflow](https://stackoverflow.com/questions/70897715/l5-swagger-swagger-php-support-for-x-logo-within-the-openapi-3-0-1-in-the-oa/70913884#70913884). 

`/**
 * @OA\Info(
 *   title="Example",
 *   version="1.0.0",
 *   x={
 *     "logo": {
 *       "url": "path_to_your_logo"
 *     }
 *   }
 * )
 */`

Will give the expected outcome - I will try to get the l5-swagger team to add an example to their documentation as the only reason I couldn't find the answer myself is because I was looking for a very specific use case that they don't currently cover in their documentation.

PR #1086, opened 2022-01-28
Update type-hinting on security property for operations.

SlvrEagle23:
We just updated versions of PhpStan, and the latest version is spitting this error back at us repeatedly: 

```
Parameter $security of attribute class OpenApi\Attributes\Get constructor expects array<string>|null, array<string, array> given.
```

Basically, the parameter I'm passing is:

```
security: [
  [
    'ApiKey': []
  ]
]
```

...which is correct as per the spec, but the hinting in the docblock is expecting an array of strings, when it's really a multidimensional array.

I think this should resolve that issue from the static typing side.

### DIFF
--- a/src/Attributes/OperationTrait.php
+++ b/src/Attributes/OperationTrait.php
@@ -11,7 +11,7 @@ use OpenApi\Generator;
 trait OperationTrait
 {
     /**
-     * @param array|null                $security
+     * @param array                     $security
      * @param Server[]                  $servers
      * @param string[]                  $tags
      * @param Parameter[]               $parameters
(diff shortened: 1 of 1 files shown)


DerManoMann:
Couldn't we then use `array<string,[]>` which would be closer to the truth?

SlvrEagle23:
@DerManoMann Sorry, my example was incorrect and I've corrected it; basically, it's a two-level-deep array, so it would actually more match what PhpStan is saying, i.e. `array<int, array<string, array>>`

DerManoMann:
Well, whatever we can do to be more specific. Happy to also merge as-is and improve later if needed.

SlvrEagle23:
@DerManoMann Yeah, I'm not 100% confident on that being the _only_ possible structure for that, that's why I just defaulted back to the safer "array" for now.

DerManoMann:
Thanks @SlvrEagle23

PR #1089, opened 2022-02-01
Fix Discriminator Attribute mapping property type hint

b-vadym:
In OpenApi\Annotations\Discriminator $mapping typehint string[]


DerManoMann:
@b-vadym I think the docblock change needs formatting. If you have dev dependencies installed you should be able to fix that by running
```shell
composer cs
```

b-vadym:
@DerManoMann cs fixed

DerManoMann:
Thanks @b-vadym

PR #1090, opened 2022-02-01
Add Style and Explode to Parameter attribute constuctor

b-vadym:
I need to set a `style: deepObject` and `expose` for the query parameter. I found no other way out than to add these two parameters to the attribute constructor.


DerManoMann:
Thanks, @b-vadym

PR #1097, opened 2022-02-03
fix(Attributes): add MediaType to RequestBody

### LINKED ISSUE #1096: How to specify OA\MediaType in OA\RequestBody using Attributes?
Using annotations, the `@OA\MediaType` annotation can be simply added as an additional unnamed parameter to `@OA\RequestBody`:

https://github.com/zircote/swagger-php/blob/dd5f1b7b84ffa86527e984b292c5223aeac5f50f/Examples/petstore.swagger.io/controllers/UserController.php#L35-L45

Using PHP Attributes, I would expect the following to work, but the `RequestBody` attribute constructor only accepts parameters `attachables`, `content`, `description` and `x`, but no additional parameters, such as `mediaType[s]`:

```php
use OpenApi\Attributes as OA;

#[OA\Post(
    path: "/user/createWithArray",
    tags: ["user"],
    summary: "Creates list of users with given input array",
    description: "",
    operationId: "createUsersWithArrayInput",
    requestBody: new OA\RequestBody(
        description: "List of user object",
        required: true,
        mediaType: new OA\MediaType(
            mediaType: "multipart/form-data",
            schema: new OA\Schema(
                type: "array",
                items: new OA\Items(ref: "#/components/schemas/User")
             )
         )
     ),
    responses: [
      new OA\Response(response: "default", description: "successful operation")
   ]
)]
 ```

**edit**: FWIW #1084 does not cover this.

caugner:
Fixes #1096.

### DIFF
--- a/src/Attributes/RequestBody.php
+++ b/src/Attributes/RequestBody.php
@@ -21,13 +21,14 @@ class RequestBody extends \OpenApi\Annotations\RequestBody
         $content = null,
         // annotation
         ?array $x = null,
-        ?array $attachables = null
+        ?array $attachables = null,
+        ?MediaType $mediaType = null,
     ) {
         parent::__construct([
                 'description' => $description ?? Generator::UNDEFINED,
                 'required' => $required ?? Generator::UNDEFINED,
                 'x' => $x ?? Generator::UNDEFINED,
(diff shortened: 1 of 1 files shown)


caugner:
Closing, because it looks like `content` should be used.

PR #1098, opened 2022-02-03
fix(Attributes): type RequestBody content

caugner:
Related to #1096.


DerManoMann:
Loooking good, just one thing - The spec states content is Map[string, [Media Type Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#mediaTypeObject)], so could we make the type hint `array<string, MediaType>|null` to match?

DerManoMann:
Oh, breaks the tests :)

Looks like content can also be `JsonContent|XmlContent` as those are shortcuts to avoid havng to type the content-type as array key all the time...

So something like
```php
    * @param array<string, MediaType>|JsonContent|XmlContent|null          $content
     */
    public function __construct(
        array|JsonContent|XmlContent|null $content = null,
 ```

Not sure what happens when you try a list of `JsonContent` like `content: [new JsonContent()]`

File: src/Attributes/RequestBody.php [hunk 1]
```diff
@@ -12,13 +12,14 @@
...
     /**
+     * @param array<string, MediaType>|JsonContent|XmlContent|null          $content
      * @param array<string,string>|null $x
      * @param Attachable[]|null         $attachables
      */
     public function __construct(
         ?string $description = null,
         ?bool $required = null,
-        $content = null,
+        ?array $content = null,
```
DerManoMann, line 22: I think this line should be 
```php
        array|JsonContent|XmlContent|null $content = null,
```

DerManoMann (changes requested):
You also will need to run php-cs-fixer, something like

```shell
> composer cs
```

src/Attributes/RequestBody.php
caugner: Union types were only introduced in PHP 8.0, but swagger-php still supports PHP 7.2, so this would not work.

caugner: Ah, nice, I wouldn't have thought of this possibility!

DerManoMann (approved):

DerManoMann:
Thanks @caugner

PR #1104, opened 2022-02-04
Includes 'example' argument in Examples attribute constructor (PHP 8.1)

abbluiz-userh:
I had to include "$example" argument to OpenApi\Attributes\Examples constructor, because the library was complaining it was missing. Here's an example where it complained:

```php
    #[OA\Get(
        path: '/api/v1/activation-codes/{id}/modules',
        summary: 'Gets activation code modules',
        operationId: 'listmodules',
        tags: ['Activation Codes'],
        parameters: [new OA\Parameter(ref: '#/components/parameters/ActivationCodeIdRequiredInPath')],
        responses: [
            new OA\Response(
                response: 200,
                description: 'Successful operation.',
                content: new OA\JsonContent(
                    ref: '#/components/schemas/ModuleListResponse',
                    examples: [new OA\Examples(
                        summary: '2 activation code modules found',
                        value: ''
                    )]
                )
            ),
            new OA\Response(response: 204, ref: '#/components/responses/SuccessfulEmptyList'),
        ]
    )]
    public function listModules(Request $request, $id)
    {
        return ContentResource::toJson(
            resources: $this->activationCodeService->findAllModules(
                id: $id,
                load: $request->load,
                paginate: $request->paginate
            ),
            message: __('One or more ActivationCode Module found.')
        );
    }
```
More specifically, this part of the code above:

```php
                    examples: [new OA\Examples(
                        summary: '2 activation code modules found',
                        value: ''
                    )]
```

After the changes of this PR, now I can simply specify the "example" field of `OA\Example()` and the error is gone:

```php
                    examples: [new OA\Examples(
                        example: 'TwoActivationCodeModulesFound',
                        summary: '2 activation code modules found',
                        value: ''
                    )]
```

### DIFF
--- a/src/Attributes/Examples.php
+++ b/src/Attributes/Examples.php
@@ -16,6 +16,7 @@ class Examples extends \OpenApi\Annotations\Examples
      * @param Attachable[]|null         $attachables
      */
     public function __construct(
+        ?string $example = null,
         ?string $summary = null,
         ?string $description = null,
         ?string $value = null,
@@ -26,6 +27,7 @@ class Examples extends \OpenApi\Annotations\Examples
         ?array $attachables = null
     ) {
         parent::__construct([
+            'example' => $example ?? Generator::UNDEFINED,
             'summary' => $summary ?? Generator::UNDEFINED,
             'description' => $description ?? Generator::UNDEFINED,
             'value' => $value ?? Generator::UNDEFINED,


DerManoMann:
I cannot see this in the specs - what is the difference between that and using value?

abbluiz-userh:
@DerManoMann I am assuming the "example" field is the "id" of the Examples. So I can reference it like `#/components/examples/IdOfTheExample`... I added because it complained of not having added that, so I thought it was missing. This did not happen when I used Annotations.

abbluiz-userh:
@DerManoMann This is the error I get when I don't include the "example" field:
```
@OA\Examples() is missing key-field: "example" in \App\Http\Controllers\ActivationCodeController->listModules() in /www/app/Http/Controllers/ActivationCodeController.php on line 221
```

DerManoMann:
Ok, maybe my phone is not the best tool to review this ;)

DerManoMann:
Well, `OpenApi\Annotations\Examples` does have the `example` property so I suppose it is fine :)

PR #1110, opened 2022-02-08
Enables more attribute targets for OpenApi\Attributes\Examples.

abbluiz-userh:
This is a suggestion. Is there a reason why Examples attribute can only be attributed to classes? In this PR I enabled the targets methods and properties, just like the OpenApi\Attributes\Schema attribute.


DerManoMann:
No real reason. I've just treid to give it some sense but there was always the chance this would happen.

DerManoMann:
Thanks @abbluiz-userh

PR #1116, opened 2022-02-10
Allow Tag and SecuritScheme to be repeatable

### LINKED ISSUE #1113: Tag and SecurityScheme attributes are not repeatable
I'm currently updating one of our code bases to use attributes instead of annotations. I've noticed a major change in that we can't define multiple tags or security schemes on a single class due to those things not having the IS_REPEATABLE flag.

The intent behind defining multiple of those things in one class is defining them in one centralized place, and then using those definitions in many classes. 

We'd place them closer to the code they're used in, but some tags don't really belong to any single class, and security schemes don't really belong to any class.

Was the non-repeatability of these two things intentional? If so, what's the suggested work around? At present I'm defining empty classes and putting one tag or security scheme above each, but that doesn't feel quite ideal.

Thanks for any guidance or information you can provide!

DerManoMann:
Fixes #1113

### DIFF
--- a/src/Attributes/SecurityScheme.php
+++ b/src/Attributes/SecurityScheme.php
@@ -8,7 +8,7 @@ namespace OpenApi\Attributes;
 
 use OpenApi\Generator;
 
-#[\Attribute(\Attribute::TARGET_CLASS)]
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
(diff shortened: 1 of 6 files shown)


DynamoGeek:
Woh, that was fast! Thanks so much for the quick response!

DerManoMann:
No problem - I'll build a release later this weekend

PR #1127, opened 2022-02-18
Only exit with code > 0 if the logged message was above notice

bcremer:
This allows to run `bin/openapi` on code that emits errors of type E_DEPRECATED.

Deprecated code should not result in a failing tool, see https://wouterj.nl/2021/11/deprecations-are-not-errors

### DIFF
--- a/src/Loggers/ConsoleLogger.php
+++ b/src/Loggers/ConsoleLogger.php
@@ -23,9 +23,6 @@ class ConsoleLogger extends AbstractLogger implements LoggerInterface
     ];
 
     /** @var bool */
-    protected $called = false;
-
-    /** @var bool */
     protected $loggedMessageAboveNotice = false;
 
     /** @var bool */
@@ -34,11 +31,6 @@ class ConsoleLogger extends AbstractLogger implements LoggerInterface
     public function __construct(bool $debug = false)
     {
         $this->debug = $debug;
-    }
-
-    public function called()
-    {
-        return $this->called;
     }
 
     public function loggedMessageAboveNotice()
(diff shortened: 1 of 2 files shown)


src/Loggers/ConsoleLogger.php
```diff
@@ -57,6 +71,10 @@ public function log($level, $message, array $context = []): void
         }
         $stop = !empty($color) ? static::COLOR_STOP : '';
 
+        if (!in_array($level, self::UNCRITICAL_LOG_LEVELS, true)) {
```
bcremer, line 74: I don't like the current naming of `UNCRITICAL_LOG_LEVELS` and `loggedError`.

Will rename to `containsMessageAboveNotive`.

bcremer: done

DerManoMann:
Is there any value in keeping the `called` property and method? The only (bad) reason is to keep track of the final processing status and that use case is now gone..

bcremer:
`called()` is still used to generate an extra blank line between log output and the command output: 
https://github.com/zircote/swagger-php/blob/4.2.9/bin/openapi#L216-L218

I would be fine with removing that "feature" to get rid of the extra property.

DerManoMann:
Me too :)

bcremer:
> Is there any value in keeping the called property and method?

And it's gone now.

DerManoMann:
Thanks @bcremer.

This logger is a strange beast. I really would like to separate the logging from the error handling / return value logic but at the time that was more than I was prepared to do ...

PR #1133, opened 2022-02-21
Fix a doc link to Annotation support

kkreft:
I'm not convinced we still promote Annotations over Attributes but dead link is not an answer.


DerManoMann:
Thanks @kkreft 

I agree with your sentiment - lets fix the link first and then think about rewording this. I think maybe a more general statement and a link to the reference itself might be better.
We just need to keep in mind that the docs use 'to annotate' to describe the process of adding meta-data to your source, irrespective of whether that is adding annotations or attributes.

Issue #1082, opened 2022-01-22
Is annotation of class constants supported in v4?


sadikinfosec:
As far as I can tell it doesn't seem to be, but I wanted to double-check. 

The use case here is there's a class constant thats an associative array with string keys and string values, effectively functioning as a makeshift string enum. The values are part of the api documentation so it seems to make the most sense to have the corresponding comment block adjacent to the class constant, but this doesn't seem to be recognized in v4.

The workaround I chose was to put it on a nearby, related class method instead, after which it seems to show up in the generated json as expected

DerManoMann:
Hmmm, it's an idea. Could you perhaps post a little example of how that would look?

DerManoMann:
Fixed via #1193

PR #1147, opened 2022-03-01
[attributes] add support for class constants

cdaguerre:
This PR : 
- adds support for attributes on class constants, see example below 
   related to https://github.com/zircote/swagger-php/issues/1082)
- adds support for the `const` keyword to attributes
- adds support for "backporting" `const` to single-valued enums in OAS versionw < 3.1 through a [dedicated processor](https://github.com/cdaguerre/swagger-php/blob/107d9b4d23f8d37186513dffe5c0434122b6ae0f/src/Processors/BackportConstants.php)

## 1. Attributes on class constants
```php
use OpenApi\Attributes as OA;

#[OA\Schema()]
class Airport
{
    #[OA\Property(property='kind')]
    public const KIND = 'Airport';
}
```

## 2. Support the `const` keyword in attributes
```php
use OpenApi\Attributes as OA;

#[OA\Schema()]
class Airport
{
    #[OA\Property(
        property: 'kind',
        type: 'const',
        const: 'Airport'
    )]
    public string $kind = 'Airport';
}
```
This ^^code is equivalent to the example above in terms of generated documentation.
## 3. Backport `const` type properties for versions < 3.1

The `const` keyword was introduced in OAS 3.1, eg.
```yaml
components:
  schemas:
    Airport:
        properties:
          kind:
            type: const
            const: Airport
```
In OAS < 3.1, the recommended way of doing this was to declare a single-valued enum instead, eg.
```yaml
components:
  schemas:
    Airport:
        properties:
          kind:
            type: enum
            enum: 
              - Airport
```
I added a [processor](https://github.com/cdaguerre/swagger-php/blob/107d9b4d23f8d37186513dffe5c0434122b6ae0f/src/Processors/BackportConstants.php) that will transform `const` to `enum` when the given OpenApi version is < 3.1.


src/Analysers/AttributeAnnotationFactory.php
```diff
@@ -74,6 +74,22 @@ public function build(\Reflector $reflector, Context $context): array
...
+
+            if ($reflector instanceof \ReflectionClass) {
+                foreach ($reflector->getReflectionConstants() as $rc) {
+                    foreach ($rc->getAttributes(Property::class) as $attribute) {
+                        $instance = $attribute->newInstance();
+                        if (Generator::isDefault($instance->property)) {
+                            $instance->property = $rc->getName();
+                        }
+                        if (Generator::isDefault($instance->type)) {
+                            $instance->type = 'const';
```
DerManoMann, line 86: As far as I can see 'const' is not a valid type in OAS?

src/Analysers/AttributeAnnotationFactory.php
```diff
@@ -74,6 +74,22 @@ public function build(\Reflector $reflector, Context $context): array
...
+                            $instance->const = [$rc->getValue()];
```
DerManoMann, line 87: Reading http://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.1.3 I would think this should be a plain value rather than an array?

src/Analysers/AttributeAnnotationFactory.php
DerManoMann: Yes, I am not arguing the const keyword, but the spec does not list `const` as a valid value for `type`: https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.1, unless I am missing something else here?

DerManoMann (changes requested):
This is an awesone PR! Thanks for taking the time to be so thorough.

If I could perhaps ask for a couple more (small) things:
* Add a little test
* I wonder if your great PR description could be used to add to the docs somewhere?
   For now the cookbook might be best I think. (If not that is also fine, I can do that later)

Other than that this looks good to go once the question about `type = 'const'` is sorted.

src/Analysers/AttributeAnnotationFactory.php
DerManoMann: I would think the `type` should be the type of the `const` value?

cdaguerre: Sorry I didn't get your point in the first place.
Would you agree that simply removing line 86 is the way to go? This leaves setting the type to the user instead of trying to infer it.

DerManoMann: I guess that would be good enough to start with.

cdaguerre:
@DerManoMann whilst trying to add a test I found out that class constant annotations are surprisingly not supported by `doctrine/annotations` (yet)... (ref. https://github.com/doctrine/annotations/pull/263) 
Also, there seem to be no tests for attributes and I'm not sure where to go from here.
I think I still makes sense to support class constants with attributes even if they're not supported with annotations but I'd like to have your opinion on this.

DerManoMann:
> class constant annotations are surprisingly not supported ...

Hmm, not sure that is relevant here as we pull docblocks ourself and stuff them into doctrine. 
To support annotations (and generally, really, sorry I didn't spot that earlier) the code should move from the attribute factory into the `ReflectionAnalyser`.

I had a quick play around and ended up with this near the bottom of `analyzeFqdn()`:
```php
        foreach ($rc->getReflectionConstants() as $constant) {
            foreach ($this->annotationFactories as $annotationFactory) {
                $definition['constants'][$constant->getName()] = $ctx = new Context([
                    'constant' => $constant->getName(),
                    'comment' => $constant->getDocComment() ?: Generator::UNDEFINED,
                    'annotations' => [],
                ], $context);
                foreach ($annotationFactory->build($constant, $ctx) as $annotation) {
                    if ($annotation instanceof Property) {
                        if (Generator::isDefault($annotation->property)) {
                            $annotation->property = $constant->getName();
                        }
                        if (Generator::isDefault($annotation->const)) {
                            $annotation->const = $constant->getValue();
                        }
                        $analysis->addAnnotation($annotation, $ctx);
                    }
                }
            }
        }
```

DerManoMann:
For testing you could add a const to the three basic.php under `tests/Fixtures/Apis/xxx` and update the shared `basic.yaml` accordingly.

DerManoMann:
@cdaguerre Are you still going to work on this? Happy to clean up if it feels like it is too much.

cdaguerre:
@DerManoMann it'd be great if you could take over ;)

DerManoMann:
No problem. It would be a shame to throw away your work.

DerManoMann:
Replaced by #1193

PR #1163, opened 2022-03-11
More ref gen

DerManoMann:
Pull in more details from the annotation/attribute docblocks.

image


DerManoMann:
Now is the time to improve the annotation / attribute docblocks.

* Class / property comments for annotations
* Class/ ctor params for attributes

DerManoMann:
I feel there could be a bit more styling for the paramerters/properties but that can come later.

Issue #1023, opened 2021-12-13
how to deprecate api in swagger-php?


omidxplimbo:
hi . i want to do deprecate some api and show in ui swagger . can i do ?

mroeling:
According to OpenApi documentation you can set the deprecated flag to true. `deprecated: true`
https://swagger.io/docs/specification/2-0/paths-and-operations/ --> "Marking as Deprecated"

DerManoMann:
Looks like `deprecated` as property only exists on `Schema` and subclasses. Not sure if that is enough or even correct...

DerManoMann:
Actually, I was wrong. All `Operation` derived annotations have a `deprecated` property too, so closing...

PR #1175, opened 2022-03-24
Added deprecated to OperationTrait

tm1000:
While migrating my code base from Docblock annotations to Attributes I noticed that `deprecated` was missing from the list os supported parameters on OperationalTrait

See also: https://github.com/zircote/swagger-php/issues/1023#issuecomment-1034146418

### DIFF
--- a/src/Attributes/OperationTrait.php
+++ b/src/Attributes/OperationTrait.php
@@ -31,10 +31,10 @@ trait OperationTrait
         ?array $parameters = null,
         ?array $responses = null,
         ?ExternalDocumentation $externalDocs = null,
+        ?bool $deprecated = null,
         // annotation
         ?array $x = null,
-        ?array $attachables = null,
-        ?bool $deprecated = null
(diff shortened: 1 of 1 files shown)


DerManoMann (changes requested):
Looking good, thanks.

Could I just ask you to move `deprecated` after `externalDocs` for both lines? All attributes have `x` and `attachables` as the last two parameters in the list.
For the parent constructor that would be actually after `tags`, I suppose.

tm1000:
Sure I just wasn't sure if something like that would be a breaking change but it seems like the intent here is to only allow named parameters

DerManoMann:
> Sure I just wasn't sure if something like that would be a breaking change but it seems like the intent here is to only allow named parameters

Yeah, its purely cosmetical, but since I took care to get them all the same it hurts to see the pattern broken :)

DerManoMann:
Thanks, @tm1000

PR #1218, opened 2022-04-29
get enum backing value

recca0120:


DerManoMann:
Finally, to fix those broken tests make the `setUp()` method look like this to skip the tests when using the `TokenAnalyser`:

```php
    public function setUp(): void
    {
        parent::setUp();
        
        if ($this->getAnalyzer() instanceof TokenAnalyser) {
            $this->markTestSkipped();
        }
    }

```

DerManoMann:
Thanks @recca0120

PR #1232, opened 2022-05-11
Docs codeblocks

DerManoMann:
Include a tab component and wrap in `codeblock` component to show annotation/attribute sample code side by side.

image


DerManoMann:
Mobile styling could use some more love and I am still [fighting](https://github.com/Jacobs63/vue3-tabs-component/issues/15) the url `#hash` update  which makes some pages jump.

PR #1233, opened 2022-05-12
Change enum type from string[] to string[]|int[]|float[]

berrugo:
At the moment you can only use strings in enum, if you use integer or floats, static code analyzers such as Psalm will trigger an error.

E.g. 
When defining an enum of integer values like the one below:
`#[OA\Property(enum: [123, 456, 789], default: 123)]`

Psalm will trigger an InvalidScalarArgument  (https://psalm.dev/docs/running_psalm/issues/InvalidScalarArgument/)
`Argument 1 of OpenApi\Attributes\Property::__construct expects array<array-key, string>|null, array{123, 456, 789} provided`

Thanks @tm1000 for pointing in this direction

### DIFF
--- a/docs/reference/annotations.md
+++ b/docs/reference/annotations.md
@@ -146,7 +146,7 @@ On top of this subset, there are extensions provided by this specification to al
 #### Properties
 ---
 <dl>
-  <dt><strong>ref</strong> : <span style="font-family: monospace;">string</span></dt>
+  <dt><strong>ref</strong> : <span style="font-family: monospace;">string|object</span></dt>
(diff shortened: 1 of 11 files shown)


DerManoMann:
I hope you didn't update the docs `.md` files manually - there is a script for that.... I usually just forget to commit the changes as the gh-pages workflow uses the same code.
OTOH, it certainly would be nice to keep the checked in markup files up-to-date...

berrugo:
I updated the docs automatically using the docs:build script in the composer file. Also the tests and linter passed

DerManoMann:
Thanks @berrugo

PR #1239, opened 2022-05-18
Allow original backing value in Enum

ginnerpeace:
A `Backed Enum` may be backed by types of int or string

But when I want to use integer in Enum, I saw like this:

image

-----

These changes can make enum use name or value specifiable.

image


DerManoMann:
Thanks @ginnerpeace

PR #1245, opened 2022-05-20
fix return type for Symfony debug class

shyim:
Stops the annoying symfony deprecation warnings that the return type is missing


DerManoMann:
Thanks @shyim

Issue #1220, opened 2022-05-02
Duplicate definition on same line (attribute)


tjveldhuizen:
Should [`\OpenApi\Annotations\AbstractAnnotation::matchNested`](https://github.com/zircote/swagger-php/blob/e6d00f990393ea69159a1c62886527e897676560/src/Annotations/AbstractAnnotation.php#L595) also check if the `$parent` class does not start with 'OpenApi\Attributes'? When I upgrade from version 4.2.11 to 4.3.0 I get the following error, which is not given when I check for the mentioned string.

```
User Warning: Multiple @OA\Response() with the same response="200":
\CompetitietypeController->cgetAction() in /app/src/ApiController/V1/Beach/CompetitietypeController.php on line 94
\CompetitietypeController->cgetAction() in /app/src/ApiController/V1/Beach/CompetitietypeController.php on line 94
```

Indeed, a warning about a duplicate definition on the same line. The line number seems to be incorrect, the reference is to line 29 in the code below (file has 50 lines)

```php
28    #[OA\Get(summary: 'Geeft een overzicht van competitietypes', tags: ['Definitietabellen'])]
29    #[OA\Response(response: 200, description: 'Success')]
30    #[Api\Parameter\Page]
31    #[Api\Parameter\Limit]
32    #[Route(methods: ['GET'])]
33    public function cgetAction(ParamFetcherInterface $paramFetcher, PaginationService $paginationService): Response
34    {
```

DerManoMann:
> Should [\OpenApi\Annotations\AbstractAnnotation::matchNested](https://github.com/zircote/swagger-php/blob/e6d00f990393ea69159a1c62886527e897676560/src/Annotations/AbstractAnnotation.php#L595) also check if the $parent class does not start with 'OpenApi\Attributes'?

Unfortunately no. Doing so will break quite some tests. Parent/child relations are always configured using annotations - attributes inherit those - so we always need to go back to annotations in order to get correct matches.

Its hard for me to tell what is wrong here without a way to reproduce this error - would it be possible for you to create a simple repository that isolates the issue? I know that some changes made here are not that compatible with framework bundles and it would be cool to be able to debug some of those issues...

tjveldhuizen:
I'll try to make a reproduction. I'm using this in a Symfony project in which I use https://github.com/nelmio/NelmioApiDocBundle. The base path is on the class level, this specific method has no extension of that base path. In 4.2.11 that was working.

[...]

PR #1279, opened 2022-07-01
Do not return annotations merged into an explicit parent in AttributeAnnotationFactory

GuilhemN:
I was looking for a solution to fix https://github.com/nelmio/NelmioApiDocBundle/issues/1993 (explanation of root issue in https://github.com/zircote/swagger-php/issues/1220#issuecomment-1116895871), and I couldn't find an easy one from NelmioApiDocBundle without duplicating a lot of code and making things more likely to break later.

I believe the best solution would be to not return annotations that are already merged in `AttributeAnnotationFactory` so that these may not be merged again later.
This should have no repercussion on zircote/swagger-php internals as attributes returned by `AttributeAnnotationFactory` are passed to `Analysis::addAnnotations` which visits recursively passed annotations and so is able to find merged attributes from their newly defined parents.

What do you think? Would that be an acceptable fix for you?

### DIFF
--- a/src/Analysers/AttributeAnnotationFactory.php
+++ b/src/Analysers/AttributeAnnotationFactory.php
@@ -120,22 +120,27 @@ class AttributeAnnotationFactory implements AnnotationFactoryInterface
             return get_class($annotation) != get_class($possibleParent)
                 && ($explicitParent || ($isAttachable && $isParentAllowed));
         };
+
+        $annotationsWithoutParent = [];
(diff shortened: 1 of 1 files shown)


DerManoMann:
I am currently looking into these issues too, but more from a 'how do I track all nested annotations' POV.
Tests pass and this brings it more in line with docblocks, so yeah, fine with me :)

DerManoMann:
Thanks @GuilhemN

PR #1039, opened 2021-12-23
Allow `nullable` next to `$ref`

DerManoMann:
Supersedes PR #921

PR #921, opened 2021-03-24
Nullable refs

raziel057:
Fix https://github.com/zircote/swagger-php/issues/918

PR #1303, opened 2022-08-28
Allow attributes' enums to be specified with an enum class string

k2tzumi:
# Proposal

When using Attribute to describe specifications, we would like to synchronize the enum property with the Enum class.

I thought it would be a good idea to have the enum value automatically expanded from the Enum class name.

```php
#[Property('status', type: 'string', enum: StatusEnum::class, example: StatusEnum::DRAFT)]
public StatusEnum status;
```

# What we dare not do

* Allowing `UnitEnum` arrays to be specified.
In PHP8.1, IDE and linter warn when calling `EnumClass::cases()` in Attribute, so it was decided not to support it at this time.
For example, here are some warnings

  * `Constant expression contains invalid operations`
  * `Cannot call abstract static method UnitEnum::cases().`

Could be a problem with the PHP version, but do you have any knowledge of this issue?

### DIFF
--- a/tests/Processors/ExpandEnumsTest.php
+++ b/tests/Processors/ExpandEnumsTest.php
@@ -84,6 +84,7 @@ class ExpandEnumsTest extends OpenApiTestCase
 
     /**
      * @requires PHP >= 8.1
+     *
      * @dataProvider expandEnumClassStringFixtures
      */
     public function testExpandEnumClassString(array $files, array $expected): void
--- a/tests/Processors/ExpandEnumsTest.php
+++ b/tests/Processors/ExpandEnumsTest.php
@@ -100,7 +100,7 @@ class ExpandEnumsTest extends OpenApiTestCase
     }
 
     /**
-     * @param list<\UnitEnum> $enums
+     * @param list<StatusEnum> $enums
      *
      * @return list<string>
      */
--- a/tests/Processors/ExpandEnumsTest.php
+++ b/tests/Processors/ExpandEnumsTest.php
@@ -7,6 +7,10 @@
 namespace OpenApi\Tests\Processors;
 
 use OpenApi\Analysers\TokenAnalyser;
+use OpenApi\Annotations\Items;
+use OpenApi\Annotations\Property as AnnotationsProperty;
+use OpenApi\Attributes\Property as AttributesProperty;
+use OpenApi\Generator;
 use OpenApi\Processors\ExpandEnums;
 use OpenApi\Tests\Fixtures\PHP\StatusEnum;
 use OpenApi\Tests\Fixtures\PHP\StatusEnumBacked;
@@ -59,5 +63,28 @@ class ExpandEnumsTest extends OpenApiTestCase
         $schema = $analysis->getSchemaForSource(StatusEnumStringBacked::class);
 
         self::assertEquals(['draft', 'published', 'archived'], $schema->enum);
+    }
+
+    /**
+     * @requires PHP >= 8.1
+     */
+    public function testExpandEnumClassString(): void
+    {
+        $analysis = $this->analysisFromFixtures(['PHP/ReferencesEnum.php']);
+        $analysis->process([new ExpandEnums()]);
+        $schemas = $analysis->getAnnotationsOfType([AnnotationsProperty::class, AttributesProperty::class, Items::class], true);
+
+        $expected = [
+            'statusEnum' => array_map(fn ($c) => $c->name, StatusEnum::cases()),
+            'statusEnumBacked' => array_map(fn ($c) => $c->value, StatusEnumBacked::cases()),
+            'statusEnumIntegerBacked' => array_map(fn ($c) => $c->value, StatusEnumIntegerBacked::cases()),
+            'statusEnumStringBacked' => array_map(fn ($c) => $c->value, StatusEnumStringBacked::cases()),
+            'statusEnums' => Generator::UNDEFINED,
+            'itemsStatusEnumStringBacked' => array_map(fn ($c) => $c->value, StatusEnumStringBacked::cases()),
+        ];
+
+        foreach ($schemas as $schema) {
+            self::assertEquals($expected[$schema->title], $schema->enum);
+        }
     }
 }
--- a/tests/Processors/ExpandEnumsTest.php
+++ b/tests/Processors/ExpandEnumsTest.php
@@ -65,23 +65,32 @@ class ExpandEnumsTest extends OpenApiTestCase
         self::assertEquals(['draft', 'published', 'archived'], $schema->enum);
     }
 
+    public function expandEnumClassStringFixtures(): iterable
+    {
+        return [
+            [
+                ['PHP/ReferencesEnum.php'],
+                [
+                    'statusEnum' => $this->convertEnumNames(StatusEnum::cases()),
+                    'statusEnumBacked' => $this->convertEnumValues(StatusEnumBacked::cases()),
+                    'statusEnumIntegerBacked' => $this->convertEnumValues(StatusEnumIntegerBacked::cases()),
+                    'statusEnumStringBacked' => $this->convertEnumValues(StatusEnumStringBacked::cases()),
+                    'statusEnums' => Generator::UNDEFINED,
+                    'itemsStatusEnumStringBacked' => $this->convertEnumValues(StatusEnumStringBacked::cases()),
+                ],
+            ],
+        ];
+    }
+
     /**
      * @requires PHP >= 8.1
+     * @dataProvider expandEnumClassStringFixtures
      */
-    public function testExpandEnumClassString(): void
+    public function testExpandEnumClassString(array $files, array $expected): void
     {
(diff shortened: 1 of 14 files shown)


DerManoMann (changes requested):
This looks quite good and the fact that there is a test makes it even better :)

The only thing I'd prefer done differently is the use of the `Util` class.
I've been trying to move away from using it and/or adding to it. There is a lot of logic spread across a lot of files and in additrion I would suspect that this doesn't work for annotations?

As an alternative the `convertEnum` code should be added to the `ExpandEnums` processor which would cover all cases in a single place. Looks like `$enum` is coming from either `Schema` or `ServerVariable`, so adding some new processing loop like this should deal with all annotations/attributes:
```
            /** @var AnnotationSchema[] $schemas */
            $schemas = $analysis->getAnnotationsOfType([AnnotationSchema::class, AttributeSchema::class, ServerVariable::class]);

            }
        }
```

k2tzumi:
Thank you.

> There is a lot of logic spread across a lot of files and in additrion I would suspect that this doesn't work for annotations?

Yes.
We wanted to use Attrube only, since Enum itself is supported by PHP 8.1 and above.

But it would be nice if it could be used for annotations as well.
Since annotations and attributes are treated as similar constructs in the source, it was necessary to review some of the annotation sources as well.

> As an alternative the convertEnum code should be added to the ExpandEnums processor which would cover all cases in a single place. Looks like $enum is coming from either Schema or ServerVariable, so adding some new processing loop like this should deal with all annotations/attributes:

I am very happy to learn that there is an expansion point!

I understood the source line to be as follows
https://github.com/zircote/swagger-php/blob/master/src/Processors/ExpandEnums.php#L29-L34

I haven't caught up with understanding the source yet, but is it like rewriting the context enum value in the above loop?

I have one question.
Will the context enum property feel like only strings are allowed and not the original array coming across?

https://github.com/zircote/swagger-php/blob/master/src/Context.php#L28

DerManoMann:
> I understood the source line to be as follows https://github.com/zircote/swagger-php/blob/master/src/Processors/ExpandEnums.php#L29-L34
> 
> I haven't caught up with understanding the source yet, but is it like rewriting the context enum value in the above loop?

Yes, you'd just need a separate loop in order to catch also `ServerVariable` which does not extend from `Schema`.
It might be an idea to break down the `ExpandEnums` code into separate methods - one for the current loop and a second one for the loop to be added.

> I have one question. Will the context enum property feel like only strings are allowed and not the original array coming across?
> 
> https://github.com/zircote/swagger-php/blob/master/src/Context.php#L28

The context `enum` is different. That will contain the enum class name in case the parsed file contains an enum, just like the `class` property contains the class name.
See: https://github.com/zircote/swagger-php/blob/master/src/Analysers/ReflectionAnalyser.php#L86

tests/UtilTest.php
```diff
@@ -79,4 +79,44 @@ public function testShorten(array $classes, array $expected): void
     {
         $this->assertEquals($expected, Util::shorten($classes));
     }
+
+    public function convertEnumsFixtures(): iterable
+    {
+        return [
+            [\OpenApi\Tests\Fixtures\PHP\StatusEnumBacked::class, [1, 2, 3]],
```
DerManoMann, line 86: Could you please use `use` statements and import those enums rather than using FQDN?

k2tzumi:
I have been following the implementation and feel that it can be handled with the following description.

```php
#[Schema(description: "backend status", example: StatusEnumBacked::DRAFT)]
enum StatusEnumBacked: int
{
    case DRAFT = 1;
    case PUBLISHED = 2;
    case ARCHIVED = 3;
}
```

```php
#[Property('status', refs: '#/components/schemas/StatusEnumBacked')]
public StatusEnumBacked status;
```

Is it universal usage to refer to it by ref as in this code?

The only problem with the way I wrote this is that I could not overwrite the nullable value.

```php
#[Property('status', refs: '#/components/schemas/StatusEnumBacked', nullable: true)]
public ?StatusEnumBacked status;
```

Would it be preferable to eliminate this problem?
If we were to have it override by the referenced property, would it be difficult to handle because of the large impact?

k2tzumi:
> The only problem with the way I wrote this is that I could not overwrite the nullable value.

There was a related Issue.
https://github.com/zircote/swagger-php/issues/918

Is there any chance that the following PR will be merge?

https://github.com/zircote/swagger-php/pull/921
https://github.com/zircote/swagger-php/pull/1039

It would be helpful to know if there are any factors that are blocking you.

DerManoMann:
> > The only problem with the way I wrote this is that I could not overwrite the nullable value.

Yes, that is an known issue that is not easy to solve.
 
> Is there any chance that the following PR will be merge?
> 
> #921 #1039

Not as they are. The 3.0.0 spec does not allow any other keys next to `$ref` and 3.1.0 only  `summary` and `description`.
Setting the OpenApi version to 3.1.0 should deal with nullable but I suspect that doesn't work in combination with `ref`.
I suggest opening a new issue for that.

> It would be helpful to know if there are any factors that are blocking you.

Time and money :)

Still waiting for the updates to the PR as requested.

DerManoMann:
On the topic of using `ref` for an enum - that is not neccesary as they `enum` typehint is enough.
```php
<?php

use OpenApi\Attributes as OAT;

#[OAT\Schema(description: "backend status", example: StatusEnumBacked::DRAFT)]
enum StatusEnumBacked: int
{
    case DRAFT = 1;
    case PUBLISHED = 2;
    case ARCHIVED = 3;
}

#[OAT\Schema]
class Model
{
    #[OAT\Property(nullable: true)]
    public ?StatusEnumBacked $status;
}
```

will generate this YAML:
```
openapi: 3.0.0
components:
  schemas:
    StatusEnumBacked:
      description: 'backend status'
      type: integer
      enum:
        - DRAFT
        - PUBLISHED
        - ARCHIVED
      example: 1
    Model:
      properties:
        status:
          $ref: '#/components/schemas/StatusEnumBacked'
      type: object
```

k2tzumi:
> Not as they are. The 3.0.0 spec does not allow any other keys next to $ref and 3.1.0 only summary and description.
Setting the OpenApi version to 3.1.0 should deal with nullable but I suspect that doesn't work in combination with ref.
I suggest opening a new issue for that.

I understood that the OpenAPI specification makes it difficult to achieve this in the first place.

Shall we resume consideration when we have more time and money..  :sweat_smile:

> Still waiting for the updates to the PR as requested.

I understand that this modification will be an effective way to define nullable enums, so I would like to proceed.
Please give me some more time.

k2tzumi:
> It might be an idea to break down the ExpandEnums code into separate methods - one for the current loop and a second one for the loop to be added.

Should I separate the Processor class( `ExpandEnums` ) itself?

With a class name like `ExpandEnumClasssString`

I am a little unsure if it will work with the Processor dependencies, but I am thinking of running `ExpandEnumClasssString` after `ExpandEnums`.

DerManoMann:
> > It might be an idea to break down the ExpandEnums code into separate methods - one for the current loop and a second one for the loop to be added.
> 
> Should I separate the Processor class( `ExpandEnums` ) itself?
> 
> With a class name like `ExpandEnumClasssString`
> 
> I am a little unsure if it will work with the Processor dependencies, but I am thinking of running `ExpandEnumClasssString` after `ExpandEnums`.

Nah, just leave it all in the same processor, just split into two separate methods.

k2tzumi:
I took your advice and modified ExpandEnums.

https://github.com/zircote/swagger-php/pull/1303/commits/820b159a880ef4139c7bf9f0a7187741b1616879

I haven't been able to implement the test, but is this what it looks like?

k2tzumi:
Tests have been implemented.
https://github.com/zircote/swagger-php/pull/1303/commits/e480ebecf8f01d0c98257eb8794037a343835089

DerManoMann:
you can run `composer cs` to make php-cs-fixer fix the code style for you.. You'll need 8.1 for that though + latest deps

DerManoMann (changes requested):

tests/Processors/ExpandEnumsTest.php
```diff
@@ -65,11 +65,12 @@ public function testExpandBackedStringEnum(): void
         self::assertEquals(['draft', 'published', 'archived'], $schema->enum);
     }
 
-    /**
-     * @requires PHP >= 8.1
-     */
     public function expandEnumClassStringFixtures(): iterable
     {
+        if (!class_exists('\\ReflectionEnum')) {
```
DerManoMann, line 70: Would PHP run this even if the test is disabled :/

k2tzumi: It may be a specification of dataProvider, but it seems to be working.
https://github.com/k2tzumi/swagger-php/runs/8181815888?check_suite_focus=true#step:5:36

DerManoMann (approved):
Some nitpicking but good enough to merge :)

DerManoMann:
Thanks @k2tzumi

Issue #1311, opened 2022-09-12
ref by class name ?


momala454:
Is it possible to reference a schema using the class name instead of the schema name ?

the following doesn't work obviously as it doesn't return something in the form of "#/components/schemas/foo" when doing Foo::class

```php
(...)
new OA\Schema(
                        properties: [new OA\Property(
                            property: 'file_profile',
                            ref: Foo::class
                        )]
                    )


namespace helloWorld;
#[\OpenApi\Attributes\Schema()
class foo
{
}
```

DerManoMann:
It's something that is not consistent yet, but using type instead of ref might work

momala454:
thanks

momala454:
With my use case, it seems to work with requestBody, but not with responses. I saw you're refactoring it in https://github.com/zircote/swagger-php/issues/1310, great !

likemusic:
@momala454  Make sure that `\MyNamespace\Class` has `OA\Schema` attribute.

momala454:
> 

it doesn't have #[\OA\Schema] but it have an extended Attribute :

```php
#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
class ApiExceptionAttribute extends \OpenApi\Attributes\Schema
{}
```

It doesn't work when I do it like this. Using "type" works only if I use directly `\OpenApi\Attributes\Schema`

likemusic:
> It doesn't work when I do it like this. Using "type" works only if I use directly \OpenApi\Attributes\Schema

@momala454 I have the same issue yesterday)

Look at here: https://github.com/zircote/swagger-php/blob/4f63dcbdd216501125870b409139ff7324f3703e/src/Processors/ExpandClasses.php#L26

Processor that expand classes for schemas do this just for `OA\Schema` and `OAT\Schema` (`true` flag at `$analysis->getAnnotationsOfType()` call) because many other annotation classes inherited from Schema class and processed separately by other processors.

![image](https://user-images.githubusercontent.com/603401/191298729-c810a392-eb04-49d4-bd3a-246074fb17b2.png)

At the moment I think that the best way to fix it would be add $exclude (array) parameter for `getAnnotationsOfType()` method to be able to extract schemas with child classes except already defined in package.

momala454:
Changing the `true` strict parameter to false in the class `ExpandClasses` is not enough to fix it

likemusic:
[...]

PR #1315, opened 2022-09-21
Fix Schema inheritance issue.

likemusic:
Fix issue described in https://github.com/zircote/swagger-php/issues/1311#issuecomment-1252435696

> > 
> 
> it doesn't have #[\OA\Schema] but it have an extended Attribute :
> 
> ```
> #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
> class ApiExceptionAttribute extends \OpenApi\Attributes\Schema
> {}
> 
> 
> #[ApiExceptionAttribute]
> class Class{}
> ```
> 
> It doesn't work when I do it like this. Using "type" works only if I use directly `\OpenApi\Attributes\Schema`

### DIFF
--- a/src/Analysis.php
+++ b/src/Analysis.php
@@ -7,7 +7,6 @@
 namespace OpenApi;
 
 use OpenApi\Annotations as OA;
-use OpenApi\Attributes as OAT;
 
 /**
  * Result of the analyser.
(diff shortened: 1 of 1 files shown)


DerManoMann:
I know this doesn't break any tests ATM, but I'll need some time thinking about this. I am pretty sure there was a reason for the explicit check... if only I could remember :/

PR #1316, opened 2022-09-24
Allow enums to be specified with an enum array

akalineskou:
I had the need to show just a subset of an enum

It showed correctly in swaggerui, but phpstan and psalm were throwing errors because I couldn't use a \BackedEnum (in my case) to pass to the $enum property

But when I tried to test it, it didnt have the expected result, so I changed the code a bit to be able to parse enums/strings/etc in an array

I used this commit as a starting place https://github.com/zircote/swagger-php/commit/48a769243b4024f3611ab9e8b137a9720d19a3bb

### DIFF
--- a/src/Processors/ExpandEnums.php
+++ b/src/Processors/ExpandEnums.php
@@ -73,20 +73,34 @@ class ExpandEnums
         ]);
 
         foreach ($schemas as $schema) {
-            if ($schema->enum !== Generator::UNDEFINED && is_string($schema->enum)) {
-                $source = $schema->enum;
-                // Convert to Enum value if it is an Enum class string
-                if (is_a($schema->enum, '\UnitEnum', true)) {
-                    $enums = [];
-                    foreach ($source::cases() as $case) {
-                        $enums[] = $case->value ?? $case->name;
-                    }
+            if (Generator::isDefault($schema->enum)) {
+                continue;
+            }
 
-                    $schema->enum = $enums;
+            if (is_string($schema->enum)) {
+                // might be enum class
+                if (is_a($schema->enum, \UnitEnum::class, true)) {
+                    $source = $schema->enum::cases();
                 } else {
-                    throw new \InvalidArgumentException("Unexpected enum value, requires specifying the Enum class string: $source");
+                    throw new \InvalidArgumentException("Unexpected enum value, requires specifying the Enum class string: $schema->enum");
+                }
+            } elseif (is_array($schema->enum)) {
+                // might be array of enum, string, int, etc...
+                $source = $schema->enum;
+            } else {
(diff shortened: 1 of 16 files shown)


src/Processors/ExpandEnums.php
```diff
@@ -73,20 +73,35 @@ private function expandSchemaEnum(Analysis $analysis): void
         ]);
 
         foreach ($schemas as $schema) {
-            if ($schema->enum !== Generator::UNDEFINED && is_string($schema->enum)) {
-                $source = $schema->enum;
-                // Convert to Enum value if it is an Enum class string
+            if ($schema->enum == Generator::UNDEFINED) {
```
DerManoMann, line 76: There is `Generator::isDefault()` for these types of check...

src/Processors/ExpandEnums.php
```diff
@@ -73,20 +73,35 @@ private function expandSchemaEnum(Analysis $analysis): void
...
+                continue;
+            }
+
+            if (is_string($schema->enum)) {
+                // might be enum class
                 if (is_a($schema->enum, '\UnitEnum', true)) {
```
DerManoMann, line 82: I think it is safe to use `\UnitEnum::class` here I think.

src/Processors/ExpandEnums.php
```diff
@@ -73,20 +73,35 @@ private function expandSchemaEnum(Analysis $analysis): void
...
-                    }
+                    $source = $schema->enum::cases();
+                } else {
+                    throw new \InvalidArgumentException("Unexpected enum value, requires specifying the Enum class string: $schema->enum");
+                }
+            } elseif (is_array($schema->enum)) {
+                // might be array of enum, string, int, etc...
+                $source = $schema->enum;
+            } else {
+                // other type (do nothing)
```
DerManoMann, line 91: Should this also throw an exception? I would think this is an unsupported value?

At least some debug logging might be nice here.

src/Processors/ExpandEnums.php
```diff
@@ -73,20 +73,35 @@ private function expandSchemaEnum(Analysis $analysis): void
...
+                continue;
+            }
 
-                    $schema->enum = $enums;
+            $enums = [];
+            foreach ($source as $enum) {
+                if (is_a($enum, '\UnitEnum')) {
```
DerManoMann, line 97: `\UnitEnum::class` again...

DerManoMann (changes requested):
Some small changes but otherwise looking good.

Apologies for making you fix things I missed in previous reviews...

DerManoMann (approved):
Lets go with that!

DerManoMann:
Thanks @akalineskou

PR #1331, opened 2022-10-19
Extend ExpandClasses processor to also consider ancestor traits

### LINKED ISSUE #1325: Merging properties from non-schema parent should include traits
Given the following:
```php
trait HasId
{
    /**
     * @OA\Property(
     *     format="int64",
     *     readOnly=true,
     * )
     *
     * @var int
     */
    public int $id;
}

abstract class Model
{
    use HasId;
}

/**
 * @OA\Schema(
 *     required={"street"},
 *     @OA\Xml(
 *         name="Address"
 *     )
 * )
 */
class Address extends Model
{
    /** @OA\Property */
    public string $street;
}
```

The `ExpandClasses` processor should merge not just direct properties properties from `Model` into `Address`, but also properties inherited from `HasId`.

Related to https://github.com/DarkaOnLine/L5-Swagger/issues/499

DerManoMann:
Fixes #1325

### DIFF
--- a/Examples/polymorphism/polymorphism-3.1.0.yaml
+++ b/Examples/polymorphism/polymorphism-3.1.0.yaml
@@ -45,6 +45,7 @@ components:
         -
           $ref: '#/components/schemas/EmployeeResponsible'
     EmployeeResponsible:
+      type: object
       allOf:
         -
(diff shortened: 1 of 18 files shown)


rickgoemans:
Any update on the release?

PR #1368, opened 2022-12-20
Allow finding child attributes in method args

ruscon:
This allows you to create custom attributes like this:

```php
use OpenApi\Attributes\Parameter;

#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY | \Attribute::TARGET_PARAMETER | \Attribute::IS_REPEATABLE)]
class QueryParameter extends Parameter
{
    public $in = 'query';
}
```

And use it in a class method like this:

```php
public function index(#[QueryParameter] ?string $name) {}
``` 

A similar solution can be done for Cookie, Header and RequestBody, I can try to implement them inside the library itself, if you want


src/Analysers/AttributeAnnotationFactory.php
```diff
@@ -53,7 +53,7 @@ public function build(\Reflector $reflector, Context $context): array
                 // also look at parameter attributes
                 foreach ($reflector->getParameters() as $rp) {
                     foreach ([OAT\Property::class, OAT\Parameter::class, OAT\PathParameter::class] as $attributeName) {
-                        foreach ($rp->getAttributes($attributeName) as $attribute) {
+                        foreach ($rp->getAttributes($attributeName, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
```
DerManoMann, line 56: Good find! I didn't know about the flag yet.

In fact, with the flag the `PathParameter::class` element can be removed from the foreach I suppose?

DerManoMann: I've been thinking about this before - hence the `PathParameter`  annotation.

I would argue that path and query are the most common cases and therefore we limit it to those two.

DerManoMann (changes requested):
For completeness it would be good to also add a QueryParameter annotation class.

Also, if you do a search for `PathParameter` in the `src` folder you'll find a lot more places where `QueryParameter` would need to be added.

Finally, it would be cool to go through the `Examples` folder and the test fixtures and change a few `Parameter` annotations to `QueryParameter` and make sure the tests still pass.

src/Analysers/AttributeAnnotationFactory.php
ruscon: > In fact, with the flag the PathParameter::class element can be removed from the foreach I suppose?

Yes, it's true :)

ruscon: @DerManoMann only one thing, for `Parameter` you have the `ParameterTrait`, but for `RequestBody` you don't.

In fact, for each base class, you need to create a trait, then it will be very easy to inherit any annotation/attribute class.

DerManoMann: Not sure I can follow you here...

DerManoMann: Ah, maybe I am. Yes, that would be handy but also quite some overhead for a non-quantifiable benefit.

ruscon: @DerManoMann About this feature.
As I said, If you want, I can try to introduce new Attributes.
In my project, I already have `QueryParameter` and `RequestBody`, 
Similar to `QueryParameter`, I can add `CookieParameter` and `HeaderParameter` attributes (without annotation class for them, because I need to extend the `Parameter` attribute class)

ruscon: > In fact, with the flag the PathParameter::class element can be removed from the foreach I suppose?

I won't do this, because this is a BC, I can just `@deprecate` it.

DerManoMann: Ah, I see. Maybe it could be like this then to avoid special treatment for `PathParameter`

```
                    foreach ([OA\Property::class, OA\Parameter::class] as $attributeName) {
                        foreach ($rp->getAttributes($attributeName, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
```

DerManoMann: @ruscon Let me know if the above works for you...

ruscon: @DerManoMann Do you mean, removing `PathParameter`?
Yes, it will work, but if somebody already uses it, you will create a breaking change.

DerManoMann: No, I mean 
```
                    foreach ([OA\Property::class, OA\Parameter::class] as $attributeName) {
                        foreach ($rp->getAttributes($attributeName, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
```

DerManoMann: Note that the alias changed from **OAT** to **OA**, so any attributes extended from the **annotation** base classes will match.

src/Attributes/ParameterTrait.php
```diff
@@ -40,7 +40,7 @@ public function __construct(
     ) {
         parent::__construct([
                 'parameter' => $parameter ?? Generator::UNDEFINED,
-                'name' => $name ?? Generator::UNDEFINED,
+                'name' => Generator::isDefault($this->name) ? $name : $this->name,
```
ruscon, line 43: Ability to change the name.
Example:
```php
public function update(
        #[CookieParameter] string $cookie,
        // with custom name
        #[HeaderParameter(name: 'test-header')] ?string $header,
    ) {}
```

src/Analysers/AttributeAnnotationFactory.php
ruscon: > Note that the alias changed from OAT to OA, so any attributes extended from the annotation base classes will match.

@DerManoMann Yep, it works with `OA`.

Some tests crashed.

```bash
1) OpenApi\Tests\Annotations\AttributesSyncTest::testPropertyCompleteness with data set "HeaderParameter" ('OpenApi\Attributes\HeaderParameter')
ReflectionException: Class "OpenApi\Annotations\HeaderParameter" does not exist
```

should I add an Annotation version for the new Attributes?

src/Attributes/ParameterTrait.php
DerManoMann: I thought I mentioned that adding all sorts of shortcut versions is a slippery slope. It's a lot of classes for a small amount of syntactic sugar.

If we add them, then  yes, there will need to be annotation versions as well.

Refactoring to merge annotations and attributes is not really possible without major BC issues.
For one thing constructors would need to match and that just doesn't add up.

ruscon: I'm just making these changes in a separate branch, to see how it behaves. In general, it turns out, but the main problem is that you need to switch to a new version of the Annotation format. [Here is the documentation](https://www.doctrine-project.org/projects/doctrine-annotations/en/1.13/custom.html#optional-constructors-with-named-parameters). So, need to change a lot of examples :D in the repo.

ruscon: When working with method parameters, this is extremely necessary, because you won't be able to name the variables the same as the name of the cookie

ruscon:
@DerManoMann Did you have a chance to check the PR?
I added annotation classes.
Locally all tests passed successfully.

DerManoMann (approved):
Looking good.

I wasn't sure we need all those specialized parameter classes, but since I started it with adding `PathParameter` I suppose it makes sense to add them all...

DerManoMann:
Thanks @ruscon

PR #1376, opened 2023-01-03
Add minProperties/maxProperties to attributes

### LINKED ISSUE #1373: [bug] Unknown named property "minProperties"
I use the schema:
```php
#[OA\Schema(schema: 'SchemaName', properties: [
    new OA\Property(property: 'property1', type: 'object', minProperties: 1, properties: [
        new OA\Property(property: 'property2', type: 'string'),
        new OA\Property(property: 'property3', type: 'string'),
    ]),
], additionalProperties: false)]
```
"minProperties" is ignored and says it's unknown.

The same, describing with the annotations works fine.

DerManoMann:
Fixes #1373


jdecool:
This PR has been tagged as a bug fix (https://github.com/zircote/swagger-php/releases/tag/4.5.4).

But it potentially introduces a BC break in applications as the constructor definition as changed, and new fields are not optional, no ?

DerManoMann:
Good point. I've never considered that, mostly because I've never used attributes without using named parameters.

The main reason for the order I inserted them was to keep them the same as they are defined in the annotations.

jdecool:
>  I've never considered that, mostly because I've never used attributes without using named parameters.

Neither me.

But I've noticed that on one project I'm currently working on

DerManoMann:
Apologies. I hope we'll do better next time.

I did add a note to the release notes, but for practical purposes that is mostly cosmetic I suppose.

PR #1381, opened 2023-01-04
Enum cleanup

### LINKED ISSUE #1337: How does the engine analyze Enums?
I commonly use int backed enums; however, for API output, the string name values are used instead of the integer, for example:
   
```
enum ExampleEnum: int implements JsonSerializable
{
    use JsonSerializeEnumNameTrait;

    case One = 1;
    case Two = 2;
    case Three = 3;
}

trait JsonSerializeEnumNameTrait
{
    public function jsonSerialize(): string
    {
        return $this->name;
    }
}
```

Due to implementing the JsonSerializable interface, and including the jsonSerialize() function in the used trait, the output in a json_encode is the case name, instead of the value.

When I use the generator, it sets the type to `integer`, and the values are displayed as the names:

My swagger looks as follows:
```
ExampleEnum integer

Enum:
[ One, Two, Three ]
````

This is interesting, as I assumed the default would be the int backed value, which is good for my case; however, I would prefer to change the type to `string`, so if I use the following attribute: 

```
#[Schema(
    type: "string"
)]
```

The output changes to:

```
ExampleEnum string

Enum:
[ 1, 2, 3 ]
```

This does the opposite of what I would have expected, it is displaying the int backed values, even though the type is set to string.

Is there a way I can get the output to display as "string", and use the name's instead of the int backed values? I'm having trouble understanding the logic of how the generator is choosing the enum values to display

DerManoMann:
Fixes #1337


DerManoMann:
Sorry, got carried away with the cleanup. Key are just `ExpandEnums` and the slight adjustments to the rules in `ExpandEnumsTest`

momala454:
This is a breaking change, and as far I understand, shouldn't version be changed to 2.0.0 intead of 1.6.0 ?

momala454:
hello @DerManoMann , what is your opinion ?

DerManoMann:
Strictly speaking that might be correct. OTOH, it could be considered a bug fix since for the cases change with this PR, the generated spec might be wrong already anyway since you can end up with int values and type `string` and vice versa.
That is why I thought a minor version bump it is justifiable.

Issue #1338, opened 2022-11-03
Marking a property as nullable


CxDevLead:
I am creating a schema as follows:

```
#[OA\Schema(    //manually create the return schema for the api response
    schema: "ResponseData",
    properties: [
        new OA\Property(
            property: "yearDisplay",
            ref: "#/components/schemas/YearDisplayModel",
            type: "object",
            nullable: true,
        ),
    ],
    type: "object",
)]
```

In the output swagger, the property 'yearDisplay' is not marked as being `null` or `YearDisplayModel`, how do I set the output to include that the property can also be null instead of the object?

DerManoMann:
This is a known issue. The spec for 3.0 doesn't allow any other properties next to `ref` and in 3.1 the only ones allowed are `summary` and `description`.

swagger-php currently just strip all other properties.

I believe the "correct" way to archive this is to use `oneOf` where you have the ref and null as possible values.  Although, I am not sure how to represent a `null` of the top of my head.

CxDevLead:
Understood, thanks, I'll add it to the descriptions.

momala454:
@DerManoMann 
How do I do "oneOf" on top of a class attribute ?

```php
class MyClass
{
    public function __construct(
        #[\OpenApi\Attributes\Property(anyOf: [new \OpenApi\Attributes\Schema(type: 'null'), new \OpenApi\Attributes\Schema(type: MyClass::class)])]
        public ?MyClass $folder,
    ) 
{}
```

the "anyOf" is ignored

DerManoMann:
Good question, never tried that. Have you tried `oneOf`?

momala454:
yes, same result, it's ignored

momala454:
@DerManoMann this is the solution i've made which seems to work.
File
vendor/zircote/swagger-php/src/Annotations/AbstractAnnotation.php

Method jsonSerialize

Modify this part
```php
// $ref
        if (isset($data->ref)) {
            // Only specific https://github.com/OAI/OpenAPI-Specification/blob/3.1.0/versions/3.1.0.md#reference-object
            $ref = ['$ref' => $data->ref];
            if ($this->_context->version == OpenApi::VERSION_3_1_0) {
                $defaultValues = get_class_vars(get_class($this));
                foreach (['summary', 'description'] as $prop) {
                    if (property_exists($this, $prop)) {
                        if ($this->{$prop} !== $defaultValues[$prop]) {
                            $ref[$prop] = $data->{$prop};
[...]

PR #1391, opened 2023-01-13
Allow a ref to be nullable

momala454:
Fixes https://github.com/zircote/swagger-php/issues/1338

It creates a oneOf with a single ref, and with nullable true

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -370,11 +370,11 @@ abstract class AbstractAnnotation implements \JsonSerializable
                 if ($this->nullable !== $defaultValues['nullable']) {
                     $ref2 = $ref;
                     $ref['oneOf'] = [
-                        $ref2
+                        $ref2,
                     ];
                     if ($this->_context->version == OpenApi::VERSION_3_1_0) {
                         $ref['oneOf'][] = ['type' => 'null'];
-                    } else {  
+                    } else {
                         $ref['nullable'] = $data->nullable;
                     }
                     unset($ref['$ref']);
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -366,6 +366,16 @@ abstract class AbstractAnnotation implements \JsonSerializable
                     }
                 }
             }
+            $defaultValues = get_class_vars(get_class($this));
+            if (property_exists($this, 'nullable') && $this->nullable === true) {
+                if ($this->nullable !== $defaultValues['nullable']) {
+                    $ref['nullable'] = $data->nullable;
+                    $ref['oneOf'] = [
+                        ['$ref' => $data->ref],
+                    ];
+                    unset($ref['$ref']);
+                }
+            }
             $data = (object) $ref;
         }
 
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -367,19 +367,13 @@ abstract class AbstractAnnotation implements \JsonSerializable
                 }
             }
             if (property_exists($this, 'nullable') && $this->nullable === true) {
-                if ($this->nullable !== $defaultValues['nullable']) {
-                    $ref2 = $ref;
-                    $ref['oneOf'] = [
-                        $ref2,
-                    ];
-                    if ($this->_context->version == OpenApi::VERSION_3_1_0) {
-                        $ref['oneOf'][] = ['type' => 'null'];
-                    } else {
-                        $ref['nullable'] = $data->nullable;
-                    }
-                    unset($ref['$ref']);
-                    unset($data->nullable);
+                $ref = ['oneOf' => [$ref]];
+                if ($this->_context->version == OpenApi::VERSION_3_1_0) {
+                    $ref['oneOf'][] = ['type' => 'null'];
+                } else {
+                    $ref['nullable'] = $data->nullable;
                 }
+                unset($data->nullable);
             }
             $data = (object) $ref;
         }
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -356,8 +356,8 @@ abstract class AbstractAnnotation implements \JsonSerializable
         if (isset($data->ref)) {
             // Only specific https://github.com/OAI/OpenAPI-Specification/blob/3.1.0/versions/3.1.0.md#reference-object
             $ref = ['$ref' => $data->ref];
-            if ($this->_context->version == OpenApi::VERSION_3_1_0) {
-                $defaultValues = get_class_vars(get_class($this));
+            $defaultValues = get_class_vars(get_class($this));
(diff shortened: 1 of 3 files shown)


momala454:
the test fails, because it looks like by default "category" on Pet is nullable, but it contains this 

```php
/**
     * @OA\Property(
     *     title="Category",
     * )
     *
     * @var Category
     */
    private $category;
```

by default, things are nullable ?

momala454:
requires https://github.com/zircote/swagger-php/pull/1397 to pass tests

momala454:
i have no idea what are those errors on the tests

DerManoMann:
I update master so a new rebase/merge should fix the remaining build issues.

momala454:
I rebased and the build is successful.
Can this PR be merged if you're ok with it ?

DerManoMann (changes requested):
Looking good I think. I don't think there is any way of `oneOf` already being set...

It would be cool if you could add a scratch test (file +spec) that tests this incl. handling of `summary` and `description`

momala454:
so should I also support summary and description the same as I support nullable ?

DerManoMann:
Both, kinda. A test for both 3.0 and 3.1 would be great.
3.1 allows those other two properties next to $ref, so I would expect those to be also in the one of.

momala454:
So for  < 3.1 I must add the possibility to add description and summary beside nullable too ? Not sure if I understood correctly

DerManoMann:
I just merged #1405 which adds a scratch test for refs in combination with `nullable` (ignored) and `description` (ignored in 3.0.0).
II would expect the yaml to change for all 4 cases with the description in the 3.1.0 case to be still present next to `$ref` inside the `oneOf`.

`refplus` endpoint in `NullRef31.yaml`:
```yaml
              schema:
                $ref: '#/components/schemas/repository'
                description: 'The repository'
```

momala454:
is it possible that something have a `$ref` AND a type ?
When i add {nullable: true, oneOf: [...]}, this code try to add `type`, but it does not exist : 

```php
if ($this->_context->version == OpenApi::VERSION_3_1_0) {
            if (isset($data->nullable)) {
                if (true === $data->nullable) {
                    $data->type = (array) $data->type;
                    $data->type[] = 'null';
                }
                unset($data->nullable);
            }
        }
```

DerManoMann:
Good point. I think unsetting `nullable` is actually the correct thing to do.

In addition there are two more things that should change:

1. The line `$defaultValues = get_class_vars(get_class($this));` is already in the method a little further up (around line 360). Moving it out of the if clause should allow to remove the one you added.
2. I think the generated yaml is not correct in terms of the nullable property.
I think the yaml should look something like (https://github.com/OpenAPITools/openapi-generator/issues/5180#issuecomment-580954887):
```yaml
OptionalAddress:
      oneOf:
      - type: 'null'
      - $ref: "#/definitions/Address"
```

momala454:
1. yes
2. this is only for 3.1 ?

DerManoMann:
I do not believe having `nullable` inside `oneOf` is legit. To me this is mostly about 3.1 support. 
The `openapi-generator` issue linked above contains some interesting variations and I would argue that adding just 3.1 support is enough until we have figured out a stable way to handle it in 3.0.

momala454:
i need it on 3.0 as I can't upgrade until swagger ui support it :/

momala454:
i updated the code for both 3.0 and 3.1.
The code style check is not successful but it doesn't say exactly what is the problem on the worker

DerManoMann:
You can fix code style issues by running `composer cs`
I'll have a look in the next few days.

src/Annotations/AbstractAnnotation.php
```diff
@@ -366,10 +366,25 @@ public function jsonSerialize()
                     }
                 }
             }
+            if (property_exists($this, 'nullable') && $this->nullable === true) {
+                if ($this->nullable !== $defaultValues['nullable']) {
+                    $ref2 = $ref;
```
DerManoMann, line 371: not sure we can rely on $ref being set here?

src/Annotations/AbstractAnnotation.php
```diff
@@ -366,10 +366,25 @@ public function jsonSerialize()
                     }
                 }
             }
+            if (property_exists($this, 'nullable') && $this->nullable === true) {
+                if ($this->nullable !== $defaultValues['nullable']) {
```
DerManoMann, line 370: if `nullable` is `true` does it matter what the default value was/is and if it differs?

DerManoMann (changes requested):
Couple questions to clean up, otherwise looking good. I wonder if we could add a non-ref element to the scratch test and make it nullable - to see what happens when `$ref` is not set.

DerManoMann:
Good stuff otherwise - looking forward to merging as this seems to be a major pain point....

src/Annotations/AbstractAnnotation.php
DerManoMann: Ah, nevermind.

src/Annotations/AbstractAnnotation.php
DerManoMann: I think the if statement on line 370 can go altogether. I do not particularly like the defaultValues stuff anyway. I'd rather is `Generator::isDefault()` although that is a different matter.

momala454: i took inspiration of some code above, I don't know if the default value is important or not. But it looks like the code changed since then. If the default value is true I suppose we still need to do the same trick so we could remove the defaultValue check

tests/Fixtures/Scratch/NullRef31.yaml
```diff
@@ -22,8 +24,10 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/repository'
                 description: 'The repository'
+                oneOf:
+                  - { $ref: '#/components/schemas/repository', description: 'The repository' }
```
DerManoMann, line 29: Not sure this duplication of the description is correct. Originally it is next to `$ref` so it does't really apply to the `null`. OTOH, we'll never have any other `oneOf` values...

If we just keep the on next to `$ref` then line 370 can be simplified to just `$ref = ['oneOf' => [$ref]];` ...

DerManoMann (changes requested):

src/Annotations/AbstractAnnotation.php
DerManoMann: It doesn't really matter what the default is - if it is `true` then its true :).  In practice the default is `Generator::DEFAULT` and `false` will also be different from the default value... so, yeah, this can go IMO.

DerManoMann (approved):
Nice work @momala454 , thanks a lot.

momala454:
thank you.
However I just remembered your message about a new test https://github.com/zircote/swagger-php/pull/1391#pullrequestreview-1294645350

momala454:
Thank you.
Can you please also take a look at my issue https://github.com/zircote/swagger-php/issues/1396
it's flooding me every time I generate my swagger file

PR #1395, opened 2023-01-19
TokenScanner: fix handling of named arguments with reserved keywords

Slamdunk:
Reserved keywords such as `class` are legit for named arguments.
Without this fix I get the following error:

```
Message: Uninitialized string offset 1

Class: ErrorException
Code: E_WARNING
File: ./vendor/zircote/swagger-php/src/Analysers/TokenScanner.php:119
```


DerManoMann:
Thanks @Slamdunk

PR #1400, opened 2023-01-29
More accurately guessing Openapi mistype line and position

matiaspub:
image

`$context->line` point to 26 line - method itself. Annotations always placed above the method/class/property/e.t.c. 
That means we have to move back(decrease `$context->line` to got line of the error in current annotation).

_I've checked other places, and can't find any place where `$context->line` could be pointed to the line of start annotation._
_I've checked older PHP versions - `ReflectionMethod::getStartLine` always returned line of the method itself, not line of the method annotation._


DerManoMann:
Thanks @matiaspub

PR #1408, opened 2023-02-09
Add MediaType requestBody support

entense:
Add type support so that x-www-form-urlencoded can be specified

### DIFF
--- a/src/Attributes/RequestBody.php
+++ b/src/Attributes/RequestBody.php
@@ -13,7 +13,7 @@ use OpenApi\Annotations as OA;
 class RequestBody extends OA\RequestBody
 {
     /**
-     * @param array<MediaType>|JsonContent|XmlContent|Attachable|null $content
+     * @param array<MediaType>|MediaType|JsonContent|XmlContent|Attachable|null $content
      * @param array<string,mixed>|null                                $x
      * @param Attachable[]|null                                       $attachables
      */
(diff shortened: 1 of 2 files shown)


src/Attributes/RequestBody.php
```diff
@@ -22,7 +22,7 @@ public function __construct(
         ?string $request = null,
         ?string $description = null,
         ?bool $required = null,
-        array|JsonContent|XmlContent|Attachable|null $content = null,
+        array|JsonContent|XmlContent|Attachable|MediaType|null $content = null,
```
DerManoMann, line 25: Could you please move `MediaType` further left, right after `array`?
ie: `array|MediaType|JsonContent|XmlContent|Attachable|null $content = null,`

DerManoMann: The same change needs to be applied to line 16 (phpdoc block). 

Once done the tests will fail since the typehint now does not match the `Annotation\RequestBody::$content` property any more.

So the annotation version of `RequestBody` needs to be updated as well (line 61):
`     * @var array<MediaType>|MediaType|JsonContent|XmlContent|Attachable`

DerManoMann (changes requested):
Couple more places to change to keep things consistent. Otherwise all good :)

DerManoMann:
Looks like some code style issues still.
You can fix those by running:

```shell
   composer cs
```

DerManoMann (changes requested):
Just fixing code style issues left and this should be good to go.

DerManoMann:
Superseded by #1426

PR #1426, opened 2023-03-16
Add MediaType as possible type to RequestBody

Malezha:
The same as #1408 but with code style fixed


DerManoMann:
Thanks @Malezha

Issue #1267, opened 2022-06-22
Hard to find out problem with invalid attribute


zim32:
Right now, invalid attribute is logged with DEBUG level.
Default logger just ignores this level.

So it is impossible to know if something whent wrong f.e.when instantiating OA\Get(...) attribute throws excpeption, it is silently dropped.
Don't know how to improve, just point this issue.

DerManoMann:
I guess this depends on whether you are using a 3rd party tool or swagger-php directly. 

You can always pass a custom `Psr\Logger` instance into the `Generator` or run `./bin/openapi` with `--debug`

DerManoMann:
I suppose https://github.com/zircote/swagger-php/blob/master/src/Analysers/AttributeAnnotationFactory.php#L52 could be conditional depending on whether the attribute class was OpenApi related or not.

Not being able to instantiate unrelated attributes (for example `JetBrains\PhpStorm\Pure` or similar) should not be considered important in the context of `swagger-php`.

zim32:
> I guess this depends on whether you are using a 3rd party tool or swagger-php directly.
> 
> You can always pass a custom `Psr\Logger` instance into the `Generator` or run `./bin/openapi` with `--debug`

Maybe it should be more explicitly mentioned in the docs. Because I need to spend quite a lot of time to figure out the problem

naitsirch:
I had the same issue and it took me several hours to find the reason why a property wasn't added to the spec. I hope my PR gets merged to prevent this kind of issues.

Issue #1162, opened 2022-03-10
Split logging and error handling


DerManoMann:
The current way of triggering errors from the logger is awkward and not very flexible.

Logging should be simplified to just do, uhm, logging.

A new mechanism needs to be added somewhere keeping track of errors. Also, how is that related to valdation?

DerManoMann:
Added to v7 wishlist

PR #1432, opened 2023-03-21
Removed error catching on invalid attributes

naitsirch:
The error catching in the `AttributeAnnotationFactory` made it hard to find the reason for "ignored" attributes. This change addresses the issues #1267 and #1162.

Instead of catching errors of non existing attribute classes, the existence of the class is checked before the attribute is tried to instantiate. This reveals invalid attribute definitions, but still errors for unknown attributes won't be shown.


DerManoMann (approved):
Probably more in line with line #57 where `newInstance()` also isn't wrapped in a `try`/`catch`

DerManoMann:
Thanks @naitsirch

PR #1436, opened 2023-04-04
Documenting Attribute references

MalekMneri:
I improved the documentation for the Post Attribute. if it is accepted, I'll try to continue with other attributes.

### DIFF
--- a/docs/reference/attributes.md
+++ b/docs/reference/attributes.md
@@ -1172,35 +1172,35 @@ In addition to this page, there are also a number of [examples](https://github.c
 ---
 <dl>
   <dt><strong>path</strong> : <span style="font-family: monospace;">string|null</span></dt>
-  <dd><p>No details available.</p></dd>
+  <dd><p>The relative or absolute path to the endpoint.</p></dd>
   <dt><strong>operationId</strong> : <span style="font-family: monospace;">string|null</span></dt>
-  <dd><p>No details available.</p></dd>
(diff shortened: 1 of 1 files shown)


MalekMneri:
Happy to be of help!
I tried updating the files in src/Attributes like this: 
image
but it won't generate after running `composer docs:gen`
Do I need to update the file tools/src/Docs/RefGenerator.php or is there another way to add a description to attributes.

DerManoMann:
The refregenerator does currently not use those docblock comments. I wasn't sure yet if it is better to focus on improving the already existing docblocks in the annotation classes and re-use those for the corresponding vars in the attribute classes.
That way we only do it once and avoid duplication. If you are happy to go down that path I'll try to find some time to update the script - something on my list already anyway...

MalekMneri:
yeah I think updating the existing annotations docblocks would be more efficient. 
I'd love to work on them after you update the script so feel free to tag me in the PR.

DerManoMann:
Try PR #1439

DerManoMann:
@MalekMneri If you think this works well enough let me know so I can merge. After that you can create a new PR with your changes and test locally.

MalekMneri:
Looks good! 
I was able to update the documentation using `composer docs:gen`
Feel free to merge it so I can continue :)

DerManoMann:
All merged - hope the syntax for refs/@see and other markup make sense :)

PR #1441, opened 2023-04-14
Better documentation for annotations/attributes

MalekMneri:
I added documentation to annotations with No details available and improved a few others.

### DIFF
--- a/docs/reference/annotations.md
+++ b/docs/reference/annotations.md
@@ -97,7 +97,7 @@ A `@OA\Request` cookie parameter.
 ---
 <dl>
   <dt><strong>in</strong></dt>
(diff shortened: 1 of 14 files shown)


DerManoMann (changes requested):
Looking good. Only thing that could be improved is that docblocks typically consist of two bits:

title/summary:
First sentence, separated by an empty line from the rest of the text.

description:
The long description.

```php
/**
 * The summary of this element.
 *
 * This is where you would go into more detail...
 */
```

MalekMneri:
I've separated the long descriptions and left the short descriptions as is.

DerManoMann:
Thanks @MalekMneri

PR #1457, opened 2023-07-28
Fix lost type 'object' when create new allOfPropertiesSchema

wubiandaxian:
When merge allOf,  a new schema "allOfPropertiesSchema" will be created, but this schema lost ”type“ property. This not feed the rules, and will cause some question when I use generate tool to create web api code which based the result json.


DerManoMann:
Looks like the test fixtures need updating. Code style also.

DerManoMann (changes requested):
Looks like there are some code style issues and existing fixtures need to be updated yet.

Fixing CS issues: `composer cs`

DerManoMann:
Closing in favour of #1489

PR #1460, opened 2023-07-31
Support Multiple Enums class-names in Schema::enum property

vad0s94:
The expression:
```php
new OA\Items(description: 'Event name', type: 'string', enum: [
      EnumFoo::class,
      EnumBar::class,
      'random-string',
])
```
will be parsed into:
```yml
items:
  description: 'Event name'
  type: string
  enum:
    - EnumFooCase1
    - EnumFooCase2
    - EnumFooCase3
    - EnumBarCase1
    - EnumBarCase2
    - random-string
```


DerManoMann:
Looks like this is breaking existing tests. Also a new test to cover this change would be great, thanks.

vad0s94:
@DerManoMann thanks! Can you update versions on Packagist?

PR #1462, opened 2023-08-01
Webhooks

qdequippe:
Version 3.1 brings "webhooks" to the specification.

- https://spec.openapis.org/oas/latest.html#openapi-document
- https://spec.openapis.org/oas/latest.html#oasWebhooks

### DIFF
--- a/docs/reference/annotations.md
+++ b/docs/reference/annotations.md
@@ -566,7 +566,7 @@ The list of values includes alternative security requirement objects that can be
 Only one of the security requirement objects need to be satisfied to authorize a request.<br />
 Individual operations can override this definition.<br />
 To make security optional, an empty security requirement `({})` can be included in the array.</p></dd>
-  <dt><strong>webhooks</strong> : <span style="font-family: monospace;">array&lt;string,PathItem&gt;</span></dt>
+  <dt><strong>webhooks</strong> : <span style="font-family: monospace;">array&lt;string,PathItem|string|class-string|object&gt;</span></dt>
   <dd><p>The available webhooks for the API.</p></dd>
 </dl>
 
--- a/docs/reference/annotations.md
+++ b/docs/reference/annotations.md
@@ -566,6 +566,8 @@ The list of values includes alternative security requirement objects that can be
 Only one of the security requirement objects need to be satisfied to authorize a request.<br />
 Individual operations can override this definition.<br />
 To make security optional, an empty security requirement `({})` can be included in the array.</p></dd>
+  <dt><strong>webhooks</strong> : <span style="font-family: monospace;">array&lt;string,PathItem&gt;</span></dt>
+  <dd><p>The available webhooks for the API.</p></dd>
 </dl>
 
 #### Reference
--- a/src/Annotations/OpenApi.php
+++ b/src/Annotations/OpenApi.php
@@ -101,7 +101,7 @@ class OpenApi extends AbstractAnnotation
     /**
      * The available webhooks for the API.
      *
-     * @var array<string,PathItem>
+     * @var array<string,PathItem|(string|class-string|object)> Array of PathItem object or Reference object
      */
     public $webhooks = Generator::UNDEFINED;
 
--- a/src/Annotations/OpenApi.php
+++ b/src/Annotations/OpenApi.php
@@ -101,7 +101,7 @@ class OpenApi extends AbstractAnnotation
     /**
      * The available webhooks for the API.
      *
-     * @var array<string,PathItem|string|class-string|object>
+     * @var array<string,PathItem>
      */
     public $webhooks = Generator::UNDEFINED;
 
@@ -113,7 +113,7 @@ class OpenApi extends AbstractAnnotation
     /**
      * @inheritdoc
      */
-    public static $_required = ['openapi', 'info', 'paths'];
+    public static $_required = ['openapi', 'info'];
 
     /**
      * @inheritdoc
--- a/src/Annotations/OpenApi.php
+++ b/src/Annotations/OpenApi.php
@@ -101,7 +101,7 @@ class OpenApi extends AbstractAnnotation
     /**
      * The available webhooks for the API.
      *
-     * @var array<string,PathItem|(string|class-string|object)> Array of PathItem object or Reference object
+     * @var array<string,PathItem|string|class-string|object>
      */
     public $webhooks = Generator::UNDEFINED;
 
--- a/src/Annotations/OpenApi.php
+++ b/src/Annotations/OpenApi.php
@@ -99,6 +99,13 @@ class OpenApi extends AbstractAnnotation
     public $externalDocs = Generator::UNDEFINED;
 
     /**
+     * The available webhooks for the API.
+     *
+     * @var array<string,PathItem>
+     */
+    public $webhooks = Generator::UNDEFINED;
+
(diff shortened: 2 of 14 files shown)


qdequippe:
It seems too easy :) am I missing something to implement? open for help :)

DerManoMann:
Nice one. It would be great if you could add some tests. Allowing a union of native types (string, etc) and `PathItem` might break things?

DerManoMann:
Also, the serialize code might need to be conditional to suppress the property for 3.0.0

qdequippe:
@DerManoMann I added test and changed types also for webhooks (only PathItem finally).

> Also, the serialize code might need to be conditional to suppress the property for 3.0.0
tests failed due to this change, any idea on how to handle this properly?

DerManoMann (changes requested):
Looking good so far.

Still some way to go to make all existing tests pass, tho. Also, it would be good to ensure that `webhook` cannot be used with 3.0.0 - something like https://github.com/zircote/swagger-php/blob/master/src/Annotations/License.php#L77

src/Annotations/OpenApi.php
```diff
@@ -106,7 +113,7 @@ class OpenApi extends AbstractAnnotation
     /**
      * @inheritdoc
      */
-    public static $_required = ['openapi', 'info', 'paths'];
+    public static $_required = ['openapi', 'info'];
```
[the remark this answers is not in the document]
qdequippe: yes, as I understand to OpenAPI Spec v3.1 `paths` is not required anymore but 

> The OpenAPI document MUST contain at least one [paths](https://spec.openapis.org/oas/latest.html#pathsObject) field, a [components](https://spec.openapis.org/oas/latest.html#oasComponents) field or a [webhooks](https://spec.openapis.org/oas/latest.html#oasWebhooks) field

So I need to find a way to add this condition to this lib.

For example https://github.com/OAI/OpenAPI-Specification/blob/3.1.0/examples/v3.1/webhook-example.yaml there is no paths only `webhooks`

DerManoMann:
Sorry for ignoring this - I am quite busy ATM with paid work :/

qdequippe:
> Sorry for ignoring this - I am quite busy ATM with paid work :/

No problem it's normal ;) I am quite busy too, it's not critical

src/Annotations/OpenApi.php
```diff
@@ -143,6 +150,18 @@ public function validate(array $stack = null, array $skip = null, string $ref =
...
         }
 
+        if ($this->openapi === self::VERSION_3_0_0 && Generator::isDefault($this->paths)) {
+            $this->_context->logger->warning('The OpenAPI document must contain paths field');
+
+            return false;
+        }
+
+        if ($this->openapi === self::VERSION_3_1_0 && Generator::isDefault($this->paths) && Generator::isDefault($this->webhooks) && Generator::isDefault($this->components)) {
+            $this->_context->logger->warning('The OpenAPI document must contain at least one paths field, a components field or a webhooks field');
```
DerManoMann, line 160: A message in similar style as above wouold be great.

Perhasps
```
At least one of 'Required @OA\Info(), @OA\Components() or @OA\Webhook() not found'
```

src/Attributes/OpenApi.php
```diff
@@ -6,17 +6,19 @@
 
 namespace OpenApi\Attributes;
 
+use OpenApi\Annotations\PathItem;
```
DerManoMann, line 9: Is this necessary? There is already a `PathItem` class in the same namespace?

DerManoMann (changes requested):
Looking pretty good! Just a few minor things...

src/Annotations/OpenApi.php
```diff
@@ -98,6 +98,13 @@ class OpenApi extends AbstractAnnotation
      */
     public $externalDocs = Generator::UNDEFINED;
 
+    /**
+     * The available webhooks for the API.
+     *
+     * @var array<string,PathItem>
```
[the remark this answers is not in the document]
qdequippe: Agree with that, I got this idea at first but I didn't know if I could add a custom annotation/attribute

src/Attributes/OpenApi.php
qdequippe: wrong use, thx

src/Annotations/OpenApi.php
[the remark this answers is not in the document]
qdequippe: Bye the way, how do you think to handle this new annotation/attribute? a new Processor `BuildWebhooks`?

docs/reference/annotations.md
```diff
@@ -566,6 +566,8 @@ The list of values includes alternative security requirement objects that can be
 Only one of the security requirement objects need to be satisfied to authorize a request.<br />
 Individual operations can override this definition.<br />
 To make security optional, an empty security requirement `({})` can be included in the array.</p></dd>
+  <dt><strong>webhooks</strong> : <span style="font-family: monospace;">array&lt;string,PathItem|string|class-string|object&gt;</span></dt>
```
DerManoMann, line 569: docs need refresh - ideally done right before merge

tests/GeneratorTest.php
```diff
@@ -38,7 +38,6 @@ public function testScan(string $sourceDir, iterable $sources): void
     public function testScanInvalidSource(): void
     {
         $this->assertOpenApiLogEntryContains('Skipping invalid source: /tmp/__swagger_php_does_not_exist__');
-        $this->assertOpenApiLogEntryContains('Required @OA\Info() not found');
```
DerManoMann, line 41: Hmm, this is surprising.

DerManoMann (changes requested):
Looks promising.

A few more things:
* `Webhook` needs to be added to the parent/nested arrays similar to other annotations in `OpenApi` (annotation)
* The attribute version should extend the annotation version. Saves on re-declaring the properties and also allows to simplify the parent/nested relationships (which are based on annotations only)

There still needs to be code to merge the Webhook data into the right place when processing the annotation tree.
Again, this is related to `$_nested`. This should probably look something like this:
```php
    public static $_nested = [
        Webhook::class => ['webhooks', 'name'],
    ];
```
From memory the first element in the array is the property this gets merged in and if there is a second value then that is used as propery lookup in the annotation (`Webhook`) to find the key to store in under.
Sounds complicated (and it kinda is, at least to understand).

src/Annotations/OpenApi.php
```diff
@@ -143,6 +151,18 @@ public function validate(array $stack = null, array $skip = null, string $ref =
...
         }
 
+        if ($this->openapi === self::VERSION_3_0_0 && Generator::isDefault($this->paths)) {
+            $this->_context->logger->warning('Required @OA\PathItem() not found');
+
+            return false;
+        }
+
+        if ($this->openapi === self::VERSION_3_1_0 && Generator::isDefault($this->paths) && Generator::isDefault($this->webhooks) && Generator::isDefault($this->components)) {
+            $this->_context->logger->warning("At least one of 'Required @OA\Info(), @OA\Components() or @OA\Webhook() not found'");
```
yceruto, line 161: It should be `@OA\PathItem` instead of `@OA\Info`

DerManoMann:
Closing in favour of #1511.

Thanks @qdequippe for doing most of the hard work.

PR #1467, opened 2023-08-21
Enum can have boolean values

momala454:


DerManoMann:
You will need to update the annotation classes too to make the tests pass...

momala454:
thanks done

src/Annotations/ServerVariable.php
```diff
@@ -27,7 +27,7 @@ class ServerVariable extends AbstractAnnotation
     /**
      * An enumeration of values to be used if the substitution options are from a limited set.
      *
-     * @var string[]|int[]|float[]|\UnitEnum[]|class-string
+     * @var string[]|int[]|float[]|\UnitEnum[]|class-string|bool[]
```
DerManoMann, line 30: Ah, just a minor thing, but one I'd like to see changed for consistency - could you move the `bool[]` adjunct to `string`,`int`,`float`, please, so the native types are all together?
 ```suggestion
     * @var string[]|int[]|float[]|bool[]|\UnitEnum[]|class-string
```

DerManoMann (changes requested):
One more find/replace that it looks good to go.

DerManoMann (approved):

DerManoMann:
Thanks @momala454

Issue #1471, opened 2023-09-01
If I set Nullable to true, the generated document will lose the title


LTaooo:
![image](https://github.com/zircote/swagger-php/assets/28125259/fb9aaf9c-f8ca-46d4-ad77-00266e2c792a)
![image](https://github.com/zircote/swagger-php/assets/28125259/9dcb7d41-d728-4582-9fc9-a86bac9e4cf0)

DerManoMann:
Is this for OpenApi 3.1.0?

LTaooo:
> Is this for OpenApi 3.1.0?

yes

PR #1472, opened 2023-09-01
When the nullable parameter of property is true, add the title parameter

LTaooo:
#1471

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -373,6 +373,9 @@ abstract class AbstractAnnotation implements \JsonSerializable
                 } else {
                     $ref['nullable'] = $data->nullable;
                 }
+                if (isset($data->title) && $data->title !== $defaultValues['title']) {
+                    $ref['title'] = $data->title;
+                }
                 unset($data->nullable);
             }
             $data = (object) $ref;


src/Annotations/AbstractAnnotation.php
```diff
@@ -373,6 +373,9 @@ public function jsonSerialize()
                 } else {
                     $ref['nullable'] = $data->nullable;
                 }
+                if (isset($data->title) && $data->title !== $defaultValues['title']) {
```
DerManoMann, line 376: Doesn't this violate the 3.1.0 spec? There is code a few lines above that explicitly allows `summary` and `description` next to `$ref`, but nothing more.

https://spec.openapis.org/oas/v3.1.0.html#schemaObject

LTaooo: YouYou are right. In OpenApi3. 1, when nullable = true, type should be array

LTaooo: It is not to add title to $ref
image

DerManoMann: Hmm, interesting. I am not sure the specs do specify that sort of thing.
If it is valid do we need to handle other properties too apart from `title`?

LTaooo: I'm not sure, maybe the description and summary can also be included. Now they are added to ref

DerManoMann: I guess we could try but there are some changes needed:
* update from master - I think I removed the `defaultValue` code
* the logic to add the title needs to be in a 3.1.0 conditional path only

LTaooo: ok

DerManoMann: Oh, I didn't mean we cannot do this, just that there are some more changes required. Or are you doing a new PR?

LTaooo: I am not doing a new PR. Do you need to merge this PR? If you need, I will reopen it.

DerManoMann: Yeah, let's reopen. Those changes have to go 😔 bro the PR.
I should have some time in the next few days for some more detailed feedback.

LTaooo: Ok, thinks

DerManoMann:
Sorry, sometimes the simplest is to make the change myself. In https://github.com/zircote/swagger-php/pull/1487 I've changed the code slightly so all other properties are preserved.

Doesn't change a lot in the tests, not sure how that works in real life though. If you could give the PR a try that would be great.

LTaooo:
> Sorry, sometimes the simplest is to make the change myself. In #1487 I've changed the code slightly so all other properties are preserved.

Thank you very much. I tested it in my project, and it seems to be working fine.
image

DerManoMann:
CLosing in favour of #1471 as that icludes the more generic handling of preserving properties.

Thanks for your help.

PR #1474, opened 2023-09-04
Fix deprecated error in PHP 8.1/8.2  

### LINKED ISSUE #1455: Deprecated: strcasecmp() - using php 8.1
Hi!,

Im getting this warning when generation schema.

`Deprecated: strcasecmp(): Passing null to parameter #2 ($string2) of type string is deprecated in../vendor/zircote/swagger-php/src/Context.php on line 286`

koriym:
This PR fixes two deprecated errors in PHP 8.1/8.2.

* `PHP Deprecated:  Creation of dynamic property Swagger\Context::{$prop} is deprecated in swagger-php/src/Context.php on line 53`
* `Deprecated: strcasecmp(): Passing null to parameter #1 ($string1) of type string is deprecated in swagger-php/src/Context.php on line 288`

No more errors when doing `composer test` in PHP 8.1/8.2.

Fixes #1455

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -136,7 +136,7 @@ abstract class AbstractAnnotation implements JsonSerializable
         foreach (static::$_blacklist as $_property) {
             unset($fields[$_property]);
         }
-        Logger::notice('Ignoring unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
+        Logger::notice('Ignoring unexpected property "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
     }
 
     /**
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -137,7 +137,6 @@ abstract class AbstractAnnotation implements JsonSerializable
             unset($fields[$_property]);
         }
         Logger::notice('Unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
-        $this->$property = $value;
     }
 
     /**
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -136,7 +136,7 @@ abstract class AbstractAnnotation implements JsonSerializable
         foreach (static::$_blacklist as $_property) {
             unset($fields[$_property]);
         }
-        Logger::notice('Unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
+        Logger::notice('Ignoring unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
     }
 
     /**
--- a/src/Annotations/Get.php
+++ b/src/Annotations/Get.php
@@ -9,7 +9,6 @@ namespace Swagger\Annotations;
 /**
  * @Annotation
  */
-#[\AllowDynamicProperties]
 class Get extends Operation
 {
     /** @inheritdoc */
--- a/src/Annotations/Get.php
+++ b/src/Annotations/Get.php
@@ -9,6 +9,7 @@ namespace Swagger\Annotations;
 /**
  * @Annotation
  */
+#[\AllowDynamicProperties]
 class Get extends Operation
 {
     /** @inheritdoc */
--- a/src/Context.php
+++ b/src/Context.php
@@ -283,7 +283,7 @@ class Context
         } else {
             $namespace = '\\'; // global namespace
         }
-        if (strcasecmp($class, $this->class) === 0) {
+        if (strcasecmp((string) $class, (string) $this->class) === 0) {
             return $namespace . $this->class;
         }
         $pos = strpos($class, '\\');
--- a/src/Context.php
+++ b/src/Context.php
@@ -33,6 +33,7 @@ namespace Swagger;
  * @property string $property
  * @property Annotations\AbstractAnnotation[] $annotations
  */
+#[\AllowDynamicProperties]
 class Context
 {
     /**
--- a/tests/AbstractAnnotationTest.php
+++ b/tests/AbstractAnnotationTest.php
@@ -18,7 +18,7 @@ class AbstractAnnotationTest extends SwaggerTestCase
 
     public function testInvalidField()
     {
-        $this->assertSwaggerLogEntryStartsWith('Unexpected field "doesnot" for @SWG\Get(), expecting');
+        $this->assertSwaggerLogEntryStartsWith('Ignoring unexpected property "doesnot" for @SWG\Get(), expecting');
         $this->parseComment('@SWG\Get(doesnot="exist")');
     }
 


DerManoMann:
Could you please explain how you got these warnings? Just running the tests on PHP 8.1 does not show any warnings for me at all.

Also, I'd rather we fix the cause of the warning instead of just allowing dynamic properties if possible.

koriym:
I did just normal `composer update;  composer test` run.
The logs are as follows.

**PHP 8.1**

<details>

```
 % composer test
> phpunit && phpcs -p --extensions=php --standard=PSR2 --error-severity=1 --warning-severity=0 ./src ./tests
PHPUnit 9.6.11 by Sebastian Bergmann and contributors.

.................................PHP Deprecated:  strcasecmp(): Passing null to parameter #1 ($string1) of type string is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 285
PHP Deprecated:  strcasecmp(): Passing null to parameter #2 ($string2) of type string is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 285
PHP Deprecated:  strcasecmp(): Passing null to parameter #1 ($string1) of type string is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 285
PHP Deprecated:  strcasecmp(): Passing null to parameter #2 ($string2) of type string is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 285
.
Deprecated: strcasecmp(): Passing null to parameter #1 ($string1) of type string is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 285

Deprecated: strcasecmp(): Passing null to parameter #2 ($string2) of type string is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 285

Deprecated: strcasecmp(): Passing null to parameter #1 ($string1) of type string is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 285
...
```
</details>

**PHP 8.2**

`Creation of dynamic property` error occurs only in PHP 8.2.

<details>

```
 % composer test
> phpunit && phpcs -p --extensions=php --standard=PSR2 --error-severity=1 --warning-severity=0 ./src ./tests
PHPUnit 9.6.11 by Sebastian Bergmann and contributors.

PHP Deprecated:  Creation of dynamic property Swagger\Context::$filename is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 249
PHP Deprecated:  Creation of dynamic property Swagger\Context::$line is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 252
PHP Deprecated:  Creation of dynamic property Swagger\Context::$method is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 256
PHP Deprecated:  Creation of dynamic property Swagger\Context::$class is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 263
PHP Deprecated:  Creation of dynamic property Swagger\Context::$namespace is deprecated in /Users/akihito/git/swagger-php/src/Context.php on line 265
PHP Deprecated:  Creation of dynamic property Swagger\Context::$comment is deprecated in /Users/akihito/git/swagger-php/src/Analyser.php on line 84
PHP Deprecated:  Creation of dynamic property Swagger\Context::$annotations is deprecated in /Users/akihito/git/swagger-php/src/Analyser.php on line 89
...
```
</details>

koriym:
The Context class assumes dynamic properties by design, and the current master branch also has a `[#[\AllowDynamicProperties]` attribute.

I think this is not an error suppression, but a design declaration, so I don't see a problem. Your thoughts?

DerManoMann:
Sorry, missed that this on the 2.x branch... I shouldn't try handling issues/PRs on my phone, I suppose...

src/Annotations/Get.php
```diff
@@ -9,6 +9,7 @@
 /**
  * @Annotation
  */
+#[\AllowDynamicProperties]
```
DerManoMann, line 12: This seems unrelated to the warnings documented?

DerManoMann (changes requested):

src/Annotations/Get.php
koriym: Without `AllowDynamicProperties` attribute, [This code](https://github.com/zircote/swagger-php/blob/2.x/src/Annotations/AbstractAnnotation.php#L140) causes a deprecate error. I thought I did not need this assignment (`$this->$property = $value;`), but I gave the attribute to minimize the changes for the behavior unchanged.

If I can get rid of this code, I don't need the `AllowDynamicProperties` attribute. (Is this unexpected property?) What should I do?

DerManoMann: Interesting. I have to admit that seem not to be able to reproduce the warnings; that would make everything a bit easier...

koriym: I am also confused. I suspect my environment and have had others try, but the error is still there. (I would expect the error to occur since the code is dealing with a dynamic propertyn in PHP 8.1.) However, the GH action I performed in another PR did not reproduce the deprecated error...

DerManoMann: :) I think you have custom error logging settings. I would have thought this should be default for phpunit :shrug: 
```
php -d error_reporting=-1 ./bin/phpunit
```

... on it.

DerManoMann: Its this test: https://github.com/zircote/swagger-php/blob/2.x/tests/AbstractAnnotationTest.php#L19

Doesn't exist on master any more, otherwise we'd see the same there...

Not quite sure what to do - technically we should probably change the code to throw an exception rather than logging and moving on; just not sure if that is too much of a BC break.
OTOH, this is the first time I've seen this issue, so I would suspect there are not a lot of cases of this happeing out in the wild. On top of that thoses properties are either ignored during serialization or bu any other OpenAPI tools.

I would lean towards not allowing invalid specs to be created, so throwing an exception might be the way forward.

koriym:
I agree with you that it is correct to throw an exception as you think. But in any case, is there any need to set invalid properties? (I pushed a [new commit](https://github.com/zircote/swagger-php/pull/1474/commits/46501392f2870af6b8b02f966c75dd8f8f2b1fbd))

src/Annotations/Get.php
koriym: I agree with you that it is correct to throw an exception as you think. But in any case, is there any need to set invalid properties? (I pushed a [new commit](https://github.com/zircote/swagger-php/pull/1474/commits/46501392f2870af6b8b02f966c75dd8f8f2b1fbd))

src/Annotations/AbstractAnnotation.php
```diff
@@ -137,7 +137,6 @@ public function __set($property, $value)
             unset($fields[$_property]);
         }
         Logger::notice('Unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
```
DerManoMann, line 139: I think we'll try this and see if anyone complains.
Could we just slightly change the log message to make it clear what is happening?
```suggestion
        Logger::notice('Ignoring unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
```

DerManoMann: I might actually pull this change into `master` too...

DerManoMann (changes requested):
One more text change and I think this is good to go.

src/Annotations/AbstractAnnotation.php
DerManoMann: https://github.com/zircote/swagger-php/pull/1476

DerManoMann:
Did you look at the PR https://github.com/zircote/swagger-php/pull/1476 ?

I actually changed field to property, but more importantly there is a test to update too (that expects this particular message)

DerManoMann:
I think there was some misunderstanding. The change to **not** set the property was fine. Only thing to change was that in my PR i also slightly changed the log message....

src/Annotations/AbstractAnnotation.php
```diff
@@ -136,8 +136,7 @@ public function __set($property, $value)
         foreach (static::$_blacklist as $_property) {
             unset($fields[$_property]);
         }
-        Logger::notice('Unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
-        $this->$property = $value;
+        Logger::notice('Ignoring unexpected field "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
```
DerManoMann, line 139:
```suggestion
        Logger::notice('Ignoring unexpected property "' . $property . '" for ' . $this->identity() . ', expecting "' . implode('", "', array_keys($fields)) . '" in ' . $this->_context);
```

... this will also require to update the test again, sorry.

src/Annotations/Get.php
DerManoMann: Could we remove this again? When not setting the property this  should not be needed.

DerManoMann (changes requested):

DerManoMann (approved):

DerManoMann:
Thanks for your patience @koriym

PR #1475, opened 2023-09-06
Migrate CI from travis to GH action for 2.x

koriym:
Test using PHP 7.2 to 8.3 with GH actions.


composer.json
```diff
@@ -34,7 +34,7 @@
     },
     "require-dev": {
         "squizlabs/php_codesniffer": ">=2.7",
-        "phpunit/phpunit": "^8 || ^9"
+        "phpunit/phpunit": "^8.5.23 || ^9"
```
DerManoMann, line 37: Do things really break with earlier 8.x versions?

koriym: With PHP 8.1 and phpunit ^8

```
./bin/phpunit                  
PHP Fatal error:  Cannot acquire reference to $GLOBALS in /Users/akihito/git/swagger-php/vendor/phpunit/phpunit/src/Util/Configuration.php on line 407
```

with PHP 8.1 and phpunit ^8.5.20

```
...............PHP Fatal error:  During inheritance of IteratorAggregate: Uncaught Return type of Symfony\Component\Finder\Finder::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

/Users/akihito/git/swagger-php/vendor/symfony/finder/Finder.php:557
```

Then I found [this advise](https://stackoverflow.com/a/70974793) at stackoverflow. And I set '^8.5.23' and see problem gone. 

At this point I didn't think I needed to explore the exact version, but after receiving your comment I tried several times changing the values. Then I found that it seems that the current state of this project is error free at `^8.5.21`.

DerManoMann:
Nice one! Thanks @koriym

PR #1490, opened 2023-10-23
Allow recursive nesting of AdditionalProperties in Attributes

rien-vroom-cquential:
Since nested AdditionalProperties have been allowed for a while through #706 also allow it through Attributes and not only with annotations.

### DIFF
--- a/src/Attributes/AdditionalProperties.php
+++ b/src/Attributes/AdditionalProperties.php
@@ -99,10 +99,9 @@ class AdditionalProperties extends \OpenApi\Annotations\AdditionalProperties
             'allOf' => $allOf ?? Generator::UNDEFINED,
             'anyOf' => $anyOf ?? Generator::UNDEFINED,
             'oneOf' => $oneOf ?? Generator::UNDEFINED,
-            'additionalProperties' => $additionalProperties ?? Generator::UNDEFINED,
             'x' => $x ?? Generator::UNDEFINED,
             'attachables' => $attachables ?? Generator::UNDEFINED,
-            'value' => $this->combine($items, $discriminator, $externalDocs, $attachables),
+            'value' => $this->combine($items, $discriminator, $externalDocs, $additionalProperties, $attachables),
         ]);
     }
 }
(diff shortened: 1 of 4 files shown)


src/Attributes/AdditionalProperties.php
```diff
@@ -98,6 +99,7 @@ public function __construct(
             'allOf' => $allOf ?? Generator::UNDEFINED,
             'anyOf' => $anyOf ?? Generator::UNDEFINED,
             'oneOf' => $oneOf ?? Generator::UNDEFINED,
+            'additionalProperties' => $additionalProperties ?? Generator::UNDEFINED,
```
DerManoMann, line 102: Have you tested this? All other nested objects are passed into the parent in this line:
```
            'value' => $this->combine($items, $discriminator, $externalDocs, $attachables),
```

DerManoMann (changes requested):
It would be great if you could add a simple example of the nesting in the `tests/Fixtures/Scratch` folder.

Something like this (still missing the nesting):
```php
// NestedAdditionalProperties.php
<?php declare(strict_types=1);

/**
 * @license Apache 2.0
 */

namespace OpenApi\Tests\Fixtures\Scratch;

use OpenApi\Attributes as OAT;

#[OAT\Schema(
    additionalProperties: new OAT\AdditionalProperties(
        type: 'string',
        schema: new OAT\Schema(
            type: 'object',
        )
    )
)]
class NestedAdditionalPropertiesModel
{
}

#[OAT\Info(
    title: 'Nested Additional Properties',
    version: '1.0'
)]
#[OAT\Get(
    path: '/api/endpoint',
    description: 'An endpoint',
    responses: [new OAT\Response(response: 200, description: 'OK', ref: NestedAdditionalPropertiesEndpoint::class)]
)]
class NestedAdditionalPropertiesEndpoint
{
}
```

Running the scratch test will create the corresponding `.yaml` file which you should check for correctness.

src/Attributes/AdditionalProperties.php
rien-vroom-cquential: For the value of additionalProperties i just copied the "OAT\Schema" style. 
I have changed it to an additional argument of the `combine()` method and the endresult is the same.
Aka it generates the same YAML file.

Change is on its way

DerManoMann (approved):
Just needs a code style fix then should be good to go.

Thanks.

DerManoMann:
Thanks @rien-vroom-cquential

Issue #1464, opened 2023-08-04
[PHP 8.2] OpenApi\\Analysis::getSchemaForSource(): Argument #1 ($fqdn) must be of type string, array given


shakaran:
PHP 8.2 with Symfony 6.3. I got this critical:

```php
Uncaught PHP Exception TypeError: \"OpenApi\\Analysis::getSchemaForSource(): 
Argument #1 ($fqdn) must be of type string, array given, called in 
/var/www/html/vendor/zircote/swagger-php/src/Processors/AugmentSchemas.php on line 106\" 
at /var/www/html/vendor/zircote/swagger-php/src/Analysis.php line 326",

"context":{"exception":
{"class":"TypeError","message":"OpenApi\\Analysis::getSchemaForSource(): 
Argument #1 ($fqdn) must be of type string, array given, called in 
/var/www/html/vendor/zircote/swagger-php/src/Processors/AugmentSchemas.php on line 106",
"code":0,"file":"/var/www/html/vendor/zircote/swagger-php/src/Analysis.php:326"}},
"level":500,"level_name":"CRITICAL"


```

DjordyKoert:
`NelmioApiDocBundle` had the same issue. I managed to fix it by updating the `api-platform/core` dependency which was used. Maybe that could also be related to your problem? 😄 

https://github.com/nelmio/NelmioApiDocBundle/pull/2126

DerManoMann:
Sounds like the root cause is unrelated to swagger-php.

Issue #1257, opened 2022-06-04
Define multiple types for a property


gmeister2:
version: 4.4.4
openapi: 3.1.0

The following Attribute code (property with mixed type: string|null):

```php
new OA\Property(
    property: 'module',
    type: 'string',
    nullable: true,
),
```

Gives this json result:

```json
"module": {
    "type": [
        "string",
        "null"
    ],
},
```

How would I setup the Attribute code to achive this json result (property with mixed type: string|float):

```json
"module": {
    "type": [
        "string",
        "float"
    ],
},
```

Tobion:
This problem is part of #885

DerManoMann:
Fixed in [5.4.2](https://github.com/zircote/swagger-php/releases/tag/5.4.2)

PR #1502, opened 2023-11-17
Feat: Change nullable & type for OpenApi v3.1.0

DjordyKoert:
Type changes to support [OpenApi v3.1.0](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0).

- Do not allow `nullable` to be set when using `OpenApi v3.1.0`
- Add array as an option to `type` property when using `OpenApi v3.1.0`

**Fixes**
- https://github.com/zircote/swagger-php/issues/1464 
  `OpenApi\\Analysis::getSchemaForSource()` method could be passed an array of types causing this `TypeError`
- https://github.com/zircote/swagger-php/issues/1257

**Related issues**
- https://github.com/zircote/swagger-php/issues/885


DjordyKoert:
I'm not exactly sure if swagger-php already correctly generatest types where nullable would normally be used for v3.1.0 because some tests started failing after adding the validation.

src/Annotations/Schema.php
```diff
@@ -483,6 +490,18 @@ public function validate(array $stack = [], array $skip = [], string $ref = '',
...
+            $this->_context->logger->error($this->identity() . ' must not have the "nullable" property when using OpenApi version ' . OpenApi::VERSION_3_1_0 . ' in ' . $this->_context);
+
+            return false;
+        }
+
+        if (is_array($this->type) && $this->_context->isVersion(OpenApi::VERSION_3_0_0)) {
+            $this->_context->logger->error($this->identity() . ' "type" must be of type string in OpenApi version ' . OpenApi::VERSION_3_0_0 . ', array given in ' . $this->_context);
+
+            return false;
+        }
```
DjordyKoert, line 503: @DerManoMann I could also get rid of this validation step for now and then simply leave this PR as is.

This would atleast allow users of this package to manually define their types according to openapi 3.1.0 standard & it would fix the `TypeError` issue https://github.com/nelmio/NelmioApiDocBundle/issues/2152, thoughts?

DjordyKoert: I have decided to remove this validation step to prevent breaking changes

DerManoMann (approved):

DerManoMann:
Thanks @DjordyKoert

DjordyKoert:
Thank you as well @DerManoMann 😄

Issue #1514, opened 2023-12-07
Serialization of data using swagger-php attributes


MattyRad:
I noticed there is a [Serializer](https://github.com/zircote/swagger-php/blob/master/src/Serializer.php) class, which does a lot of deserialization, but not a lot of serialization.

It would be really nice to serialize an object directly from swagger-php attributes using a function that is provided by the library. For example: 

```php
use OpenApi\Attributes as OpenApi;

class Sample
{
    #[OpenApi\Property(
        property: 'greeting',
    )]
    public function getGreeting(): string
    {
        return 'hello world';
    }
}

class BasicsSerializeTest extends \PHPUnit\Framework\TestCase
{
    /** @test */
    public function basic_openapi_serialization(): void
    {
        $sample = new Sample;

        $serialized = Api::openapiSerialize($sample);

        $this->assertEquals(['greeting' => 'hello world'], $serialized);
    }
}
```

We currently have the function `openapiSerialize` implemented, and it has been great!

> (`openapiSerialize` as we have it is quite large and unwieldy, hence posting the test and not the implementation).

Serializing from swagger-php attributes ensures that the API documentation *necessarily* matches the API response- i.e. you cannot add/adjust a response format without also adding/adjusting a swagger-php attribute.

I am planning on creating a repo which provides `openapiSerialize` and explaining why it's valuable, but if there is interest in me making a PR to this repo on the Serializer class (instead of a separate repo), then please let me know and I will make a PR.

Thanks for the excellent library!

DerManoMann:
There is an alternative way which is implemented here: https://github.com/DerManoMann/openapi-verifier

Basically it takesan API response and validates it against the spec. The code takes advantage of the fact that the spec is a JSON schema and therefore the API response data can be validated against the schema.

The project needs some love, though :{

PR #1515, opened 2023-12-07
Add public static serialization function to Serializer

MattyRad:
https://github.com/zircote/swagger-php/issues/1514

Ok, so it's not _pretty_... probably not even very efficient... but it **is** extremely useful! If there's any advice on how to get it fixed up and part of the repo, please let me know. Otherwise feel free to close, and I will provide this functionality in a separate package.

### DIFF
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -60,6 +60,166 @@ class Serializer
         OA\XmlContent::class,
     ];
 
+    private const GETTER_PREFIX = 'get';
+
+    /**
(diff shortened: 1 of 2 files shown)


DerManoMann:
I can see see how this could be useful, although it seems limited as it looks at models/schemas only.
Not sure it is specific enough to justify adding it to swagger-php itself.

MattyRad:
Makes sense. No problem, it can live in its own repo. https://github.com/MattyRad/openapi-serialize

Mind if I add the repo to the [related-projects list](https://github.com/zircote/swagger-php/blob/master/docs/related-projects.md)?

Issue #1528, opened 2024-01-05
Cannot overwrite generated schema


DjordyKoert:
I encountered an issue while trying to create a fix for our generated enums https://github.com/nelmio/NelmioApiDocBundle/issues/2177

While doing this I noticed that the `AugmentProperties` processor makes sure to set the `oneOf` property whenever a `ref` is set on a `nullable === true` schema.
https://github.com/zircote/swagger-php/blob/a70a5dc5db26183c86f589ff3d76fb11e141fc58/src/Processors/AugmentProperties.php#L172-L185

I wanted to overwrite this by creating a new processor which would revert this ONLY for schemas based on enums. See implementation:
```php
<?php

declare(strict_types=1);

namespace Nelmio\ApiDocBundle\Processors;

use OpenApi\Analysis;
use OpenApi\Generator;
use OpenApi\Processors\ProcessorInterface;
use OpenApi\Annotations as OA;

final class EnumOneOfToRefProcessor implements ProcessorInterface
{
    public function __invoke(Analysis $analysis): void
    {
        if (\PHP_VERSION_ID < 80100) {
            return;
        }

        $schemas = $analysis->openapi?->components->schemas;

        foreach ($schemas as $schema) {
            if (!is_array($properties = $schema->properties)) {
                continue;
            }

            foreach ($properties as $property) {
                if ($property->nullable !== true) {
                    continue;
                }

                if (!is_array($property->oneOf)) {
                    continue;
                }

                if (count($property->oneOf) !== 1) {
                    continue;
                }

                if (!$refSchema = $this->schemaForRef($schemas, $ref = $property->oneOf[0]->ref)) {
                    continue;
                }

                if (!$class = $analysis->getContext($refSchema)?->class) {
                    continue;
                }

                if (!enum_exists($class)) {
                    continue;
                }

                $property->ref = $ref;
                $property->oneOf = Generator::UNDEFINED;
            }
        }
    }

    /**
     * Find schema for the given ref.
     */
    private function schemaForRef(array $schemas, string $ref): ?OA\Schema
    {
        foreach ($schemas as $schema) {
            if (OA\Components::ref($schema) === $ref) {
                return $schema;
            }
        }

        return null;
    }
}
```

[...]

PR #1529, opened 2024-01-06
fix: undefined type property

DjordyKoert:
Related to https://github.com/zircote/swagger-php/issues/1528

Using the old nullable notation combined with `anyOf` & `allOf` does not currently add the valid `null` type as an option.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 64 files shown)


DerManoMann (approved):
I think this is fine for now; in the long run moving all this code into a processor would be a good thing IMO.

I fixed the CS and phpstan issues on `master` - once this branch is updated I am happy to merge...

DjordyKoert:
@DerManoMann done! 😄

DerManoMann:
Thanks @DjordyKoert

PR #1531, opened 2024-01-08
feat: move OpenApi 3.1.0 conversions to processor

DjordyKoert:
related to https://github.com/zircote/swagger-php/issues/1528

Moved logic to processor:
- Processing of `nullable` property for 3.1.0
- Processing of `minimum` &  `exclusiveMinimum` property for 3.1.0
- Processing of `maximum` &  `exclusiveMaximum` property for 3.1.0

### DIFF
--- a/tests/Fixtures/Scratch/NullRef31.yaml
+++ b/tests/Fixtures/Scratch/NullRef31.yaml
@@ -23,11 +23,12 @@ paths:
           description: 'Ref plus response'
           content:
             application/json:
+              example: 'Some example about the content'
               schema:
                 oneOf:
-                  - { $ref: '#/components/schemas/repository', description: 'The repository' }
+                  - { $ref: '#/components/schemas/repository' }
                   - { type: 'null' }
-                description: 'The repository'
+                description: 'Will either respond with something or null'
 components:
   schemas:
     repository: {  }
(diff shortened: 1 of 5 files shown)


tests/Fixtures/Scratch/NullRef31.yaml
```diff
@@ -23,11 +23,12 @@ paths:
           description: 'Ref plus response'
           content:
             application/json:
+              example: 'Some example about the content'
               schema:
                 oneOf:
-                  - { $ref: '#/components/schemas/repository', description: 'The repository' }
```
DjordyKoert, line 28: https://github.com/DjordyKoert/swagger-php/blob/2eedbb3a9693762720a7aff34d6a2af8303bc593/src/Annotations/AbstractAnnotation.php#L358

`$ref` no longer exists after going through the OpenApi31Processor because it gets set to `Generator::UNDEFINED` which makes it no longer copy the description.

I have another work in progress branch to test if a new annotation `Reference` can be implemented for this with instead. https://github.com/DjordyKoert/swagger-php/blob/feat/reference-object-annotation/src/Annotations/Reference.php

DerManoMann:
Closing due to inactivity... I feel this might also be better done via a serializer pipeline, similar to processors.

PR #1543, opened 2024-01-30
use @deprecated phpdoc to augment properties

cdaguerre:


DerManoMann:
Looks reasonable. Would you be able to add a simple scratch test example for this, too?

DerManoMann:
Thanks @cdaguerre

PR #1542, opened 2024-01-30
[feat] Added resolving request body name based on class name

dkulyk:
This PR added filling request name for RequestBody based on class name.

PR #1544, opened 2024-01-30
feat: Added resolving references of components (not only schemas)

dkulyk:
Includes https://github.com/zircote/swagger-php/pull/1542

Added resolving components like header, requestBody and so. 
Berole only schemas can be resolved by AugmentRefs

### DIFF
--- a/src/Processors/AugmentRefs.php
+++ b/src/Processors/AugmentRefs.php
@@ -66,7 +66,7 @@ class AugmentRefs implements ProcessorInterface
                 // check if we have a schema for this
                 if ($refSchema = $analysis->getSchemaForSource($annotation->ref)) {
                     $annotation->ref = OA\Components::ref($refSchema);
-                } elseif ($refSchema = $analysis->getAnnotationForSource($annotation->ref, get_class($annotation))) {
+                } elseif ($refAnnotation = $analysis->getAnnotationForSource($annotation->ref, get_class($annotation))) {
                     $annotation->ref = OA\Components::ref($refSchema);
                 }
             }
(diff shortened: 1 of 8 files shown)


tests/Fixtures/Request.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @OA\RequestBody
+ */
+class Request
+{
+    /**
+     * @OA\Post(
+     *     path="/",
+     *     @OA\RequestBody(ref=OpenApi\Tests\Fixtures\Request::class),
```
DerManoMann, line 21: Interesting - I didn't know this would work!

DerManoMann (changes requested):
Looking nice!

A couple nits that would be great to fix. 
Sorry for taking so long to review.

src/Processors/AugmentRefs.php
```diff
@@ -66,6 +66,8 @@ protected function resolveFQCNRefs(Analysis $analysis): void
                 // check if we have a schema for this
                 if ($refSchema = $analysis->getSchemaForSource($annotation->ref)) {
                     $annotation->ref = OA\Components::ref($refSchema);
+                } elseif ($refAnnotation = $analysis->getAnnotationForSource($annotation->ref, get_class($annotation))) {
+                    $annotation->ref = OA\Components::ref($refSchema);
```
DerManoMann, line 70: I think the variable name needs to be changed here too.

DerManoMann (changes requested):
Last bit...

DerManoMann (approved):

DerManoMann:
Thanks @dkulyk

Issue #1553, opened 2024-03-05
Supporting integer ranges from php docs


bnowak:
Hello,

Would be nice to have support of https://phpstan.org/writing-php-code/phpdoc-types#integer-ranges for more accurate schemas. I'm working on PR, which will be ready for review soon.

I open this ticket to start any potential discussions here.

bnowak:
solved and merged :)

PR #1554, opened 2024-03-05
integer ranges support

bnowak:
Th PR address https://github.com/zircote/swagger-php/issues/1553 topic


DerManoMann:
Thanks @bnowak

PR #1563, opened 2024-03-25
[4.9] Default openapi version in `Context` to 3.0.0

### LINKED ISSUE #1562: [4.8.7] Version is only available reliably for validation and serialization
Since 4.8.7 [NelmioApiDocBundle](https://github.com/nelmio/NelmioApiDocBundle) is throwing exceptions when generating documentation. 

I am not sure which change caused this or how it could be fixed (inside swagger-php or Nelmio).

https://github.com/nelmio/NelmioApiDocBundle/actions/runs/8283136995/job/23048455451

Some phpunit context:
```
1) Nelmio\ApiDocBundle\Tests\Command\DumpCommandTest::testYaml
RuntimeException: Version is only available reliably for validation and serialization

/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/zircote/swagger-php/src/Context.php:124
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/zircote/swagger-php/src/Annotations/AbstractAnnotation.php:175
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/zircote/swagger-php/src/Annotations/AbstractAnnotation.php:394
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/zircote/swagger-php/src/Annotations/AbstractAnnotation.php:[28](https://github.com/nelmio/NelmioApiDocBundle/actions/runs/8283136995/job/23048455451#step:9:29)2
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/zircote/swagger-php/src/Annotations/AbstractAnnotation.php:270
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/src/Render/Yaml/YamlOpenApiRenderer.php:30
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/src/Render/RenderOpenApi.php:81
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/src/Command/DumpCommand.php:84
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/symfony/console/Command/Command.php:[29](https://github.com/nelmio/NelmioApiDocBundle/actions/runs/8283136995/job/23048455451#step:9:30)8
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/symfony/console/Tester/CommandTester.php:74
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/tests/Command/DumpCommandTest.php:110
/home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/tests/Command/DumpCommandTest.php:44
phpvfscomposer:///home/runner/work/NelmioApiDocBundle/NelmioApiDocBundle/vendor/phpunit/phpunit/phpunit:106
```

DerManoMann:
Fixes #1562 

This PR also (re-)enables the deprecation trigger in `Context:detect()`. The problems the changes around the version check illustrate how fragile the whole concept of `Context` is and not caring about where/what is in the context is asking for trouble.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -46,11 +46,6 @@ parameters:
 			path: src/Annotations/Schema.php
 
 		-
-			message: "#^Ternary operator condition is always true\\.$#"
-			count: 1
-			path: src/Context.php
-
-		-
 			message: "#^If condition is always true\\.$#"
 			count: 1
 			path: src/Generator.php
--- a/src/Analysis.php
+++ b/src/Analysis.php
@@ -72,6 +72,8 @@ class Analysis
 
     public function addAnnotation(object $annotation, Context $context): void
     {
+        assert(!Generator::isDefault($context->version));
+
         if ($this->annotations->contains($annotation)) {
             return;
         }
@@ -369,8 +371,7 @@ class Analysis
             return $context;
         }
 
-        // Weird, did you use the addAnnotation/addAnnotations methods?
-        throw new \Exception('Annotation has no context');
+        throw new \RuntimeException('Annotation has no context - did you use addAnnotation()/addAnnotations()');
     }
 
     /**
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -166,16 +166,6 @@ abstract class AbstractAnnotation implements \JsonSerializable
     }
 
     /**
-     * Check if one of the given version numbers matches the current OpenAPI version.
-     *
-     * @param string|array $versions One or more version numbers
-     */
-    public function isOpenApiVersion($versions): bool
-    {
-        return $this->_context->isVersion($versions);
-    }
-
-    /**
      * Merge given annotations to their mapped properties configured in static::$_nested.
      *
      * Annotations that couldn't be merged are added to the _unmerged array.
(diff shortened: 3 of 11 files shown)


gnutix:
I'm not sure I understand how the bug happens and why this PR would fix it. From my POV, it seems like you had inline some logic in a method, and now you've done the opposite. But I don't see a "change". Could you explain it to me, I'm curious.

DerManoMann:
> I'm not sure I understand how the bug happens and why this PR would fix it. From my POV, it seems like you had some logic inlined in a method, and now you've reverted that inlinment. But I don't see a "change" in the logic / flow of code itself. Could you explain it to me? I'm puzzled.

The difference with using `isVersion` is that there is a check if `version` is checked at all. Comparing `context->version` directly does not have that check and doesn't care if the version is set at all.

I've removed the check and context will now default to `3.0.0` if no version is set. This is still not ideal IMO, but a safer version of how it was before.

Since the syncing of the 'final' value of the version string happens in `Generator::generate`, when creating `Context` instances outside the default flow it might be good to set the expected version on the context (just like any other context property) provided the target version is known/available at the time.

PR #1569, opened 2024-04-10
Add example schema-query-parameter-attributes processor

beeyev:
Same as `schema-query-parameter` #783 but with attributes support


DerManoMann (approved):
:+1:

PR #1575, opened 2024-04-19
Update `jsonSerialize` return

AurelienPillevesse:
When I execute my test with the last version of packages + PHP 8.3, I have warnings like this one : 

> 1x: Method "JsonSerializable::jsonSerialize()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "OpenApi\Annotations\AbstractAnnotation" now to avoid errors or add an explicit @return annotation to suppress this message.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -64,7 +64,7 @@ return (new PhpCsFixer\Config())
         'nullable_type_declaration_for_default_null_value' => true,
 
         'no_empty_phpdoc' => true,
-        'no_superfluous_phpdoc_tags' => true,
+        'no_superfluous_phpdoc_tags' => ['allow_mixed' => true],
         'phpdoc_align' => true,
         'general_phpdoc_tag_rename' => true,
         'phpdoc_inline_tag_normalizer' => true,
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -284,6 +284,9 @@ abstract class AbstractAnnotation implements \JsonSerializable
         return $properties;
     }
 
+    /**
+     * @return mixed
+     */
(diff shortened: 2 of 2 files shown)


DerManoMann:
Looks like you might have to put all changes into a single PR.

Also, tool did you use to get this warning? I was under the impression that the attribute would take care of this issue.

AurelienPillevesse:
I use : 
- `symfony` : 7.0
- `phpunit/phpunit` : 9.6.19
- `nelmio/api-doc-bundle` : 4.25.3

DerManoMann:
So is the warning from PHP itself or from a static analyser or similar?

AurelienPillevesse:
It happens when I run my tests but it seems to be a PHP warning no ?

AurelienPillevesse:
What about bumping the php version to 8.1 in my commit to make all tests pass (or in another one) because 7.x versions are not maintained anymore (and everything will be green for this PR)

DerManoMann:
It would be good to get more details as I cannot see those warnings running the swagger-php tests itself.
It would be great if you could provide perhaps a screenshot or the full output from your tests.

deluxetom:
I started to see the same warning messages with PHP 8.3 (using symfony 7 and NelmioApiDoc bundle)

AurelienPillevesse:
Added phpdoc in the `Abstract` class, will be enough I think

deluxetom (approved):

DerManoMann:
Sorry, got kind of confused about which PR does what. As you can see from my PR there is also a change in the cs-fixer config required. Happy for you to update your PR or use mine.

DerManoMann:
Thanks @AurelienPillevesse

AurelienPillevesse:
@DerManoMann do you think it's possible to release a new tag to fix this issue ?

PR #1576, opened 2024-04-19
Update `jsonSerialize` return (`Operation`)

AurelienPillevesse:
When I execute my test with the last version of packages + PHP 8.3, I have this warning : 

> 1x: Method "JsonSerializable::jsonSerialize()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "OpenApi\Annotations\Operation" now to avoid errors or add an explicit @return annotation to suppress this message.


AurelienPillevesse:
All changes in one PR, see : https://github.com/zircote/swagger-php/pull/1575

PR #1549, opened 2024-02-09
[v5]

DerManoMann:
Fixes #1540 

Looking at [packagist stats](https://packagist.org/packages/zircote/swagger-php/php-stats#4) this will affect about 1.5% of users. Moving to 8.1 would bump that number up to 12.5% which seems a bit much in one go.

Plan is to move forward to v5 reasonably soon after 8.4 is out.

Issue #1540, opened 2024-01-29
Bump PHP dep from 7.2 to 7.4 [v5]


DerManoMann:
Time to move on!?

Useful features:
* property typehints
* arrow functions

Other things to consider:
* bump symfony/finder to ^5.0
* bump symfony/yaml to ^5.0
* bump friendsofphp/php-cs-fixer (dev/suggest) to ^3.48
* bump doctrine/annotations (dev/suggest) to ^2.0
* bump phpunit/phpunit (dev) to ^9.0

Bump swagger-php version to 5

Add disclaimer about annotations being deprecated -> https://github.com/doctrine/annotations/pull/486

Remove all code marked deprecated

Add `rector` to manage refactoring and code upgrades.

Re-implement `TokenScanner` using `nikic/php-parser`

Plan is to move forward to v5 reasonably soon after 8.4 is out.

DjordyKoert:
May I ask why this package will not move to php 8 (or php 8.1 because php 8.0 is already EOL).

I got some valid feedback on  https://github.com/nelmio/NelmioApiDocBundle/pull/2171 which is why I am suggesting this. I am also currently considering moving NelmioApiDocBundle to php 8.1 https://github.com/nelmio/NelmioApiDocBundle/pull/2171#discussion_r1478458499

DerManoMann:
I think it is just a conservative approach. I do understand some of the reasons for doing it, but without a solid technical reason it still feels a bit artificial.
But then, its a step in the right direction and the issue was created for this type of discussion :)

DjordyKoert:
# Union types (on top of typed property from PHP 7.4)

A reason for moving to PHP 8 would be because of the support for union types. Currently a lot of annotations have a `@var`comment with a bunch of different possible types.
https://github.com/zircote/swagger-php/blob/bdee7f5a9216ce103ba2c953c1c43c4a3e139e4c/src/Annotations/Response.php#L69

Currently you have written your own "type parser" (correct me if I'm wrong), this could instead be handled by PHP itself (except for arrays, php has no way to handle generics :)). https://github.com/zircote/swagger-php/blob/bdee7f5a9216ce103ba2c953c1c43c4a3e139e4c/src/Annotations/AbstractAnnotation.php#L680

This also comes with a con though at this current time, because of the `Generator::UNDEFINED` constant which is currently used everywhere as a default. This would require every property to also be typed with the `string` type which could cause confusion for users of this package who might think any kind of string can be used as a value for that property. 

[...]

PR #1577, opened 2024-04-19
Bump php version

AurelienPillevesse:
Bump PHP version to last maintained one : https://www.php.net/supported-versions.php

(I can remove CI for 7.x versions in this PR or in another one)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         }
     },
     "require": {
-        "php": ">=7.2",
+        "php": ">=8.1",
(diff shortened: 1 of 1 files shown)


DerManoMann:
I think this would be a bit drastic, considering that I cannot seem to replicate the issue at all running the swagger-php test suite.

Also, this would be a huge bump. There are plans to bump, but only to 7.4 as a first setp (#1549). There is also more context/ discussion around that in #1540 .

PR #1585, opened 2024-05-01
Replace current processor array with a pipeline class

DerManoMann:
Deprecates all processor related methods on `Generator` and provides simpler, more powerful alternatives in the new `Pipeline` class that now internally wraps processors.

Also deprecates the unused `ProcessorInterface`.


mstegmeyer:
This causes #1603 and is a breaking change 😢

PR #1589, opened 2024-05-03
Document missing version option for Generator::scan()

winternet-studio:


DerManoMann:
Thanks @winternet-studio

PR #1590, opened 2024-05-03
Mention doctrine annotations must be installed manually

winternet-studio:


DerManoMann:
Thanks @winternet-studio

PR #1601, opened 2024-06-05
Collect unrelated annotations/attributes in `Context::other`

### LINKED ISSUE #1595: Accessing non-OA property attributes in processor
I'm using swagger-php with a versioned symfony API, and am using [JMS\Serializer](https://github.com/schmittjoh/JMSSerializerBundle) attributes to handle the serialization itself. 

I'm converting the docblock annotations over to attributes, but have encountered an issue that I'm not sure how to resolve. 

Previously, I could have a property such as this:
```php
/**
 * @OA\Property(
 *     description="Some property as an example",
 * ),
 * @JMS\Expose(),
 * @JMS\Since("1.2"),
 * @JMS\Until("1.3")
 */
protected $someProperty;
```

which I could run through a custom processor, grab the JMS annotations from the property `_context->annotations`, and then remove the properties that don't match the target spec version. 

However since converting to attributes, it seems as though the `AttributeAnnotationFactory` will ignore any attributes that don't extend `OA\AbstractAnnotation`, leaving the processor unable to see any non-OA attributes. Before, `_context->annotations` contained both OA and non-OA annotations.

Do you have any recommendation on how to support this without duplicating the versioning information in something like `attachables`, or would you be open to somehow exposing a properties attributes in _context?

Thanks in advance!

DerManoMann:
Fixes #1595

### DIFF
--- a/src/Analysers/AttributeAnnotationFactory.php
+++ b/src/Analysers/AttributeAnnotationFactory.php
@@ -47,9 +47,14 @@ class AttributeAnnotationFactory implements AnnotationFactoryInterface
                     $instance = $attribute->newInstance();
                     if ($instance instanceof OA\AbstractAnnotation) {
                         $annotations[] = $instance;
+                    } else {
+                        if ($context->is('other') === false) {
+                            $context->other = [];
+                        }
+                        $context->other[] = $instance;
                     }
                 } else {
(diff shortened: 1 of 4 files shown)


brettgoss (approved):
Looks good to me, thank you!

PR #1604, opened 2024-06-06
Fix nullable processor call

### LINKED ISSUE #1603: First argument of the setProcessorPipeline must be of type OpenApi\Pipeline, null given
Hello,

After upgrade to version 4.9.3 we have exception in Generator::scan method:
``OpenApi\Generator::setProcessorPipeline(): Argument #1 ($processor) must be of type OpenApi\Pipeline, null given``

This exception doesn't occur in the version 4.9.2.

Best regards

mstegmeyer:
fixes #1603 

The options `processors` was nullable before, now this throws an error, because the parameters of the generator setter are not nullable. Also, the set method for `setProcessorPipline` was executed twice, which makes the first call redundant. 

Now not provided options for `processor` and `processors` should work, with `processor` having priority over `processors`.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -286,7 +286,7 @@ class Generator
         return $this->processorPipeline->walk($walker);
     }
 
-    public function setProcessorPipeline(Pipeline $processor): Generator
+    public function setProcessorPipeline(?Pipeline $processor): Generator
     {
         $this->processorPipeline = $processor;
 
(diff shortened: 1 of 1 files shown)


mstegmeyer:
@DerManoMann please have a look, you introduced the breaking change 😉

KminekMatej:
+1 , happenned to me as well

DerManoMann (approved):
Nice catch, thank you.

Issue #1419, opened 2023-02-28
Wrong docs generation for php8 promoted properties.


zamaldev:
Using php 8 promoted properties, I expect the doccomment to be parsed the same as for the basic class properties. But it actually looks like they are overridden by method arguments. As far as I'm concerned, the only reason I want to use elevated properties is the linter hints (and one more, but it's affected by the class extension).
I tried to investigate on my own but was unsuccessful.
Maybe someone also wants to use PP like me and this could be a nice fix.
In my opinion, the problem is overriding because ReflectionClass->getProperties() returns two values with the correct doc comments.

Code:
```php
#[Schema()]
class NameValue
{
    /**
     * Property name.
     *
     * @var string
     */
    #[Property()]
    public string $name = '';

    public function __construct(
        /**
         * Property value.
         *
         * @var string
         */
        #[Property()]
        public string $value = '',
    ) {
    }
}
```

Expected result:
```json
            "NameValue": {
                "properties": {
                    "value": {
                        "description": "Property value.",
                        "type": "string",
                        "nullable": false
                    },
                    "name": {
                        "description": "Property name.",
                        "type": "string"
                    }
                },
                "type": "object"
            },
```

Actual result:
```json
            "NameValue": {
                "properties": {
                    "value": {
                        "type": "string",
                        "nullable": false
                    },
                    "name": {
                        "description": "Property name.",
                        "type": "string"
                    }
                },
                "type": "object"
            },
```

DerManoMann:
Yep, that could be better.

Also the `"nullable": false` is kinda redundant I suppose...

zamaldinov28:
And also one more bug founded here. While there is no fix yet, I tried to move all the comments to property description attribute parameter, and found that `property` property ignored.

Class:
```php
#[Schema()]
class NameValue
{
    /**
     * Property name.
     *
     * @var string
     */
    #[Property(property: 'name_override')]
    public string $name = '';

    public function __construct(
        /**
         * Property value.
         *
         * @var string
         */
        #[Property(property: 'value_override')]
        public string $value = '',
    ) {
    }
}
```

Expected result:
```json
            "NameValue": {
                "properties": {
                    "value_override": {
                        "type": "string",
                        "nullable": false
                    },
                    "name_override": {
                        "description": "Property name.",
                        "type": "string"
                    }
                },
                "type": "object"
            },
```

Actual result:
```json
            "NameValue": {
                "properties": {
                    "value": {
                        "type": "string",
[...]

PR #1619, opened 2024-07-02
Fix promoted properties sharing context

rasmuscnielsen:
A reproduction of the bug described in my comment here:  https://github.com/zircote/swagger-php/issues/1419#issuecomment-2200163118

Added regression test + fixed it with a clone of context. Tests passing. However - suggestions for alternative fixes are very welcome.

### DIFF
--- a/src/Analysers/AttributeAnnotationFactory.php
+++ b/src/Analysers/AttributeAnnotationFactory.php
@@ -80,6 +80,9 @@ class AttributeAnnotationFactory implements AnnotationFactoryInterface
                                 $instance->nullable = $nullable ?: Generator::UNDEFINED;
 
                                 if ($rp->isPromoted()) {
+                                    // ensure each property has its own context
+                                    $instance->_context = new Context([], $instance->_context);
+
                                     // promoted parameter - docblock is available via class/property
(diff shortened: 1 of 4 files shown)


src/Analysers/AttributeAnnotationFactory.php
```diff
@@ -80,6 +80,9 @@ public function build(\Reflector $reflector, Context $context): array
                                 $instance->nullable = $nullable ?: Generator::UNDEFINED;
 
                                 if ($rp->isPromoted()) {
+                                    // ensure each property has its own context
+                                    $instance->_context = clone $instance->_context;
```
DerManoMann, line 84: Not so keen on `clone` - could we use this instead?
```suggestion
                                    $instance->_context = new Context([], $instance->_context);
```

DerManoMann (changes requested):
Small change, but otherwise I suppose this is ok.

src/Analysers/AttributeAnnotationFactory.php
rasmuscnielsen: Fixed :-)

DerManoMann:
Thanks @rasmuscnielsen

PR #1646, opened 2024-08-27
Allow "processor" to be passed as scan options

pauljohnston2009:
Fixes issue with parsing 'processor' config option, it would always fall through to `new Pipeline($config['processors'])` which would throw as its null

```
TypeError: OpenApi\Pipeline::__construct(): Argument #1 ($pipes) must be of type array, null given
```

current code worked when passing 'processors', but removed support for 'processor' option.

https://github.com/zircote/swagger-php/commit/2d983ce67b9eb7e18403ae7bc5e765f8ce7b8d56

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -444,7 +444,7 @@ class Generator
             ];
 
         $processorPipeline = $config['processor'] ??
-            $config['processors'] ? new Pipeline($config['processors']) : null;
+            ($config['processors'] ? new Pipeline($config['processors']) : null);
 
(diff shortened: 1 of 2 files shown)


DerManoMann:
Let me fix CS issues on master first...

DerManoMann:
@pauljohnston2009 CS checks on master should now be stable - hopefully a rebase and code-style fix run will make this pass now...

DerManoMann:
Superseded by #1660 

Thanks for kicking off this improvement, @pauljohnston2009

PR #1647, opened 2024-08-27
allow unused tags to be kept

pauljohnston2009:
The changes in https://github.com/zircote/swagger-php/pull/1625 mean that we cannot have descriptive tags that do not have a reference to them from an operation. It is nice to allow this so we can have pages dedicated to be informative only (and allow external links directly to that section in the page).

eg. tags like

* introduction
* setup
* how-to

### DIFF
--- a/src/Processors/AugmentTags.php
+++ b/src/Processors/AugmentTags.php
@@ -17,16 +17,19 @@ class AugmentTags implements ProcessorInterface
 {
 
     /** @var array<string> */
-    protected $unusedTagsToKeepWhitelist = [];
+    protected $whitelist;
 
-    public function __construct(array $unusedTagsToKeepWhitelist = [])
+    public function __construct(array $whitelist = [])
     {
-        $this->unusedTagsToKeepWhitelist = $unusedTagsToKeepWhitelist;
+        $this->whitelist = $whitelist;
     }
 
-    public function setUnusedTagsToKeepWhitelist(array $unusedTagsToKeepWhitelist): AugmentTags
+    /**
+     * Whitelist tags to keep even if not used. <code>*</code> may be used to keep all unused.
+     */
+    public function setWhitelist(array $whitelist): AugmentTags
     {
-        $this->unusedTagsToKeepWhitelist = $unusedTagsToKeepWhitelist;
+        $this->whitelist = $whitelist;
 
         return $this;
     }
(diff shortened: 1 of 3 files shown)


src/Processors/AugmentTags.php
```diff
@@ -15,6 +15,22 @@
  */
 class AugmentTags implements ProcessorInterface
 {
+
+    /** @var array<string>  */
+    protected $unusedTagsToKeepWhitelist = [];
```
DerManoMann, line 20:
```suggestion
    protected $unusedTagsToKeepWhitelist;
```
I think the other processors only intialize in the constructor and that should be enough.

src/Processors/AugmentTags.php
```diff
@@ -15,6 +15,22 @@
...
+
+    public function __construct(array $unusedTagsToKeepWhitelist = [])
```
DerManoMann, line 22:
```suggestion
    public function __construct(array $whitelist = [])
```

It think this should be meaningful enough? This class is all about tags, after all.

src/Processors/AugmentTags.php
```diff
@@ -48,8 +65,10 @@ public function __invoke(Analysis $analysis)
             }
         }
 
+        // remove tags that we don't want to keep (defaults to all unused tags)
+        $tagsToKeep = array_merge($usedTagNames, $this->unusedTagsToKeepWhitelist);
```
DerManoMann, line 69: Could there be a special case of `'*'` which would keep all tags?

src/Processors/AugmentTags.php
```diff
@@ -15,6 +15,22 @@
...
+
+    /** @var array<string>  */
+    protected $unusedTagsToKeepWhitelist = [];
+
+    public function __construct(array $unusedTagsToKeepWhitelist = [])
+    {
+        $this->unusedTagsToKeepWhitelist = $unusedTagsToKeepWhitelist;
+    }
+
+    public function setUnusedTagsToKeepWhitelist(array $unusedTagsToKeepWhitelist): AugmentTags
```
DerManoMann, line 27:
```suggestion
    /**
     * Whitelist tags to keep even if not used. <code>*</code> may be used to keep all unused.
     */
    public function setUnusedTagsToKeepWhitelist(array $unusedTagsToKeepWhitelist): AugmentTags
```

The docblock on the `setXXX()` method is used to auto-generate the processor reference, so adding something here would be good.

DerManoMann (changes requested):
Looking good, just some nits and arguing about naming :)

src/Processors/AugmentTags.php
pauljohnston2009: based it off CleanUnusedComponents, makes sense to only do it in one spot though.

DerManoMann (approved):

DerManoMann:
Thanks @pauljohnston2009

PR #1651, opened 2024-09-21
Prefer attributes, deprecate annotations

fulldecent:


fulldecent:
This adds a deprecation notice so we can soon update to only supported PHP versions.

---

For reference, [here are the supported versions of PHP](https://www.php.net/supported-versions.php).

Please note that future versions of PHP will have longer support cycles (just for the security support, not the active support), however 7.4 and even 8.0 are specifically excluded.

This is a clear signal to the community not to use 7.4 and 8.0.

README.md
```diff
@@ -4,11 +4,12 @@
...
-[doctrine annotations](https://www.doctrine-project.org/projects/annotations.html) (optional as of version 4.8; if required the `doctrine/annotations` library must be installed in addition to swagger.php).
-or [PHP attributes](https://www.php.net/manual/en/language.attributes.overview.php).
+Generate interactive [OpenAPI](https://www.openapis.org) documentation for your RESTful API using [PHP attributes](https://www.php.net/manual/en/language.attributes.overview.php) (preferred) or 
+[doctrine annotations](https://www.doctrine-project.org/projects/annotations.html) (requires additional `doctrine/annotations` library).
 
-For a full list of supported annotations, please have look at the [`OpenApi\Annotations` namespace](src/Annotations) or the [documentation website](https://zircote.github.io/swagger-php/guide/annotations.html).
+See the [documentation website](https://zircote.github.io/swagger-php/guide/attributes.html) for supported attributes and annotations.
+
+Annotations, and support for PHP versions lower than 8.1 are deprecated and will be removed in a future release of swagger-php.
```
DerManoMann, line 12: Not quite sure about this one, in particular referencing a specific PHP version.
There are already two mentions of 8.1 as the min. required PHP version for attributes, so this doesn't really add anything.

`may be removed` is strong enough here, seeing that annotations are still widely used, IMO. If this changes faster than expected we can still change it.


```suggestion
Annotations are deprecated and may be removed in a future release of swagger-php.
```

DerManoMann (changes requested):

README.md
fulldecent: Thank you, baked in

README.md
```diff
@@ -19,6 +20,8 @@ For a full list of supported annotations, please have look at the [`OpenApi\Anno
 - Exceptional error reporting (with hints, context)
 - As of PHP 8.1 all annotations are also available as PHP attributes
 
+
```
DerManoMann, line 23: oh, another nit: why these extra lines?

DerManoMann (changes requested):
Sorry to be a pain...

README.md
fulldecent: ugggh prettier, that's why :-(

fulldecent:
Squash rebase, fix whitespace 

@DerManoMann

README.md
fulldecent: did rebase squash should be fixed now

DerManoMann (approved):
I do not particularly like the `(preferred)` - the library doesn't really care ;) I suppose it is the more modern approach :shrimp:

DerManoMann:
Thanks @fulldecent

PR #1662, opened 2024-10-07
Update generator.md

Kiina:
I just spend way too long to figure out why the example throws errors. Since you can't call new Analysis without a context and TokenAnalyser is the old way of doing things, I updated it so the  example works.

Maybe the signature for 

```php 
public function __construct(array $annotations = [], ?Context $context = null)
```
in Analysis.php should get updated since it calls addAnnotations, which doesn't allow context to be null, so I don't quite see why it allows it in the constructor since it will just crash anyway?

### DIFF
--- a/docs/reference/generator.md
+++ b/docs/reference/generator.md
@@ -56,13 +56,15 @@ $logger = new \Psr\Log\NullLogger();
 $processors = [/* my processors */];
 $finder = \Symfony\Component\Finder\Finder::create()->files()->name('*.php')->in(__DIR__);
 
+$context = new OpenApi\Context();
+
 $openapi = (new \OpenApi\Generator($logger))
             ->setProcessorPipeline(new \OpenApi\Pipeline($processors))
             ->setAliases(['MY' => 'My\Annotations'])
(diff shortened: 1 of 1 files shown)


Kiina:
Also noticed that it says:

```
processors: null|array                    Defaults to `Analysis::processors()`.
```

But Analysis::processors() has been removed in v4

DerManoMann:
Fair point. That page was originally the migration doc for switching to using the `Generator` class. I think there are a few more things to fix, so thanks for the reminder.

Kiina:
Ah yeah I've tried to implement some processors and the docs seemed a bit confusing. Some stuff only seems to be documented in the examples. I tried to extend the default pipeline with a new processor and it seems like this is only to be found in one of the examples:

https://github.com/zircote/swagger-php/blob/master/Examples/processors/schema-query-parameter/scan.php#L26

Maybe the examples should be in the header on the documentation to make them easier to find. Would've probably made it less confusing than looking through the docs and the existing Processors which use the deprecated ProcessorInterface.

DerManoMann:
Thanks, @Kiina

Issue #1668, opened 2024-10-27
Attribute behaves differently than annotation in case of DocBlockDescriptions


kminek:
This works:

```
@OA\Schema(type="string", description=null)
```

this dont:

```
`#[OA\Schema(type: 'string', description:null)]`
```

DerManoMann:
A bit more detail might be good. 'doesn't work's could be anything...

Best guess is you are facing this issue: https://zircote.github.io/swagger-php/guide/faq.html#warning-required-oa-info-not-found

kminek:
@DerManoMann right. here is my use case. i have a enum class annotated with attribute:

```

/**
 * Some docblock that should not be exposed in API docs.
 */
#[OA\Schema(type: 'string')]
enum Language: string
{
...
}
```

When API docs are generated, also enum class docblock is parsed and included in generated docs. However as this docblock contains internal info which should not be exposed i tried to set `description` to `null`:

```
/**
 * Some docblock that should not be exposed in API docs.
 */
#[OA\Schema(type: 'string', description: null)]
enum Language: string
{
...
}
```

But still the docblock is included in API docs.

DerManoMann:
Ah, yes, sorry. Was reading this on my phone and didn't scroll right :/

The attributes do not use `Generator::UNDEFINED`, so null is not really considered a valid value. Have you tried an empty string? Not sure of the top of my head how the code evaluates whether a string is set or not.
I suppose an ugly 'fix' would be to set the value to `Generator::UNDEFINED`, although that does not look very nice :{

DerManoMann:
I wonder if for attributes all properties typed `?string` could treat empty string the same as `null`... hmmm

kminek:
> I wonder if for attributes all properties typed `?string` could treat empty string the same as `null`... hmmm

dont really know the internals of this library but as a library consumer i would expect same behavior no matter what "annotating interface" i am using

DerManoMann:
Fair enough - will do some more digging then ;)

DerManoMann:
Fixed via #1886

PR #1674, opened 2024-11-22
Treat empty description same as `Generator::UNDEFINED`

DerManoMann:
Related to https://github.com/zircote/swagger-php/issues/1668


DerManoMann:
This rule should apply to all string values taken from docblocks.

Also, documentation should be updated.

DerManoMann:
Superseded by #1846

PR #1677, opened 2024-11-29
Merge Sample fixtures into Scratch tests

DerManoMann:


composer.json
```diff
@@ -64,6 +64,9 @@
         "phpunit/phpunit": ">=8",
         "vimeo/psalm": "^4.23"
     },
+    "conflict": {
+        "symfony/process": ">=6, <6.4.14"
```
umpirsky, line 68: @DerManoMann Why do we have this?

PR #1683, opened 2025-01-03
Expand on Response annotation warning message.

kevinkopf:
If a `description` or `ref` is missing in `Response` annotation, this improves the warning message, telling where exactly the missing attribute for the annotation is, saving up one's debugging time.

### DIFF
--- a/src/Annotations/Response.php
+++ b/src/Annotations/Response.php
@@ -121,7 +121,7 @@ class Response extends AbstractAnnotation
         $valid = parent::validate($stack, $skip, $ref, $context);
 
         if (Generator::isDefault($this->description) && Generator::isDefault($this->ref)) {
-            $this->_context->logger->warning($this->identity() . ' One of description or ref is required');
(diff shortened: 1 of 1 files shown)


kevinkopf:
An example of how it would look in a Symfony project:

image

DerManoMann:
I like the idea, however it might be good to handle all sorts of contexts - similar to https://github.com/zircote/swagger-php/blob/master/src/Processors/OperationId.php#L54-L72

In fact, there enums are missing too!

I suppose the `Context` class could have a method that contstructs these strings? Something like `sourceId(): string` or similar.

kevinkopf:
I went over the code and noticed that `Context` already has a method called `getDebugLocation`, which basically does what we need.

I updated my PR to reflect the change. Attaching an example message: 
image

I don't have time to go over all the warning/error messages, I'll update them one by one if/when I encounter them.

DerManoMann:
Oops

DerManoMann (approved):

DerManoMann:
Thanks @kevinkopf

PR #1689, opened 2025-01-08
fix: bracket correction in readme

ghoshrahul:
In this generator [document,](https://zircote.github.io/swagger-php/reference/generator.html#the-openapi-generator-class) the example for generator does not work, due to bracket placed incorrectly.

This change is to fix that example.


DerManoMann (approved):

DerManoMann:
Thanks @ghoshrahul

PR #1692, opened 2025-01-09
Initialize Generator::$context

### LINKED ISSUE #1693: Typed static property OpenApi\Generator::$context must not be accessed before initialization
I generate my routes with this in Laravel:

```php
<?php

use Illuminate\Support\Facades\Route as RouteFacade;
use OpenApi\Annotations\Operation;

$class = MyController::class;
$routeGroupPrefix = '/my-route';

$attributes = (new ReflectionClass($class))
    ->getMethod('__invoke')
    ->getAttributes();

foreach ($attributes as $attribute) {
    $definition = $attribute->newInstance();
    if ($definition instanceof Operation) {
        $method = strtoupper($definition->method);
        $uri = substr($definition->path, strlen($routeGroupPrefix));
        return RouteFacade::addRoute($method, $uri, $class);
    }
}
```

On `$definition = $attribute->newInstance();` an attribute instance will be created, and the Generator is not used before.

The error output is this:
```
artisan route:clear

   Error 

  Typed static property OpenApi\Generator::$context must not be accessed before initialization

  at vendor/zircote/swagger-php/src/Annotations/AbstractAnnotation.php:101
     97▕     {
     98▕         if (isset($properties['_context'])) {
     99▕             $this->_context = $properties['_context'];
    100▕             unset($properties['_context']);
  ➜ 101▕         } elseif (Generator::$context) {
    102▕             $this->_context = Generator::$context;
    103▕         } else {
    104▕             $this->_context = new Context(['generated' => true]);
    105▕         }

      +1 vendor frames 

  2   [internal]:0
      OpenApi\Attributes\JsonContent::__construct("object")
```

Can you initialize the `$context` with `null`? Or do you have a better approach?

brpauwels:
Initialize `Generator::$context` to avoid a `Typed static property OpenApi\Generator::$context must not be accessed before initialization` error when `Generator::$context` is called from the `AbstractAnnotation` constructor (line 101).

Fixes #1693

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -27,7 +27,7 @@ class Generator
     /**
      * Allows Annotation classes to know the context of the annotation that is being processed.
      */
-    public static ?Context $context;
+    public static ?Context $context = null;
 
     /** @var string Magic value to differentiate between null and undefined. */
     public const UNDEFINED = '@OA\Generator::UNDEFINED🙈';


DrIsmaileAbdallah:
i have same error this morning ! thank you @brpauwels !

DerManoMann (approved):

PR #1713, opened 2025-02-21
feat: MetaStorm Integration

xepozz:
Here is MetaStorm integration. Get more about MetaStorm:
- [Source](https://github.com/xepozz/meta-storm-idea-plugin)
- [Marketplace](https://plugins.jetbrains.com/plugin/26121-meta-storm/)

I'm not a big fan of writing open api manually, but sometimes I should do it. Last time I modified the attributes notation I've noticed that there are not suggestions for query/path/header/cookie in "in" parameter, no suggestions about media type.
I've added some to help other users do not a typo. 

One more thing is global collection `GLOBAL:media-types` will be available soon, currently it's absent it released versions.

### DIFF
--- a/.meta-storm.xml
+++ b/.meta-storm.xml
@@ -1,0 +1,39 @@
+<meta-storm xmlns="meta-storm">
+    <definitions>
+        <classConstructor class="\OpenApi\Attributes\Property" argument="0">
+            <collection name="zircote/swagger-php:types"/>
+        </classConstructor>
+        <classConstructor class="\OpenApi\Attributes\ParameterTrait" argument="3">
+            <collection name="zircote/swagger-php:path-in"/>
+        </classConstructor>
+        <classConstructor class="\OpenApi\Attributes\SecurityScheme" argument="1">
+            <collection name="zircote/swagger-php:path-in"/>
+        </classConstructor>
+        <classConstructor class="\OpenApi\Attributes\MediaType" argument="0">
+            <collection name="GLOBAL:media-types"/>
+        </classConstructor>
+        <classMethod class="\OpenApi\Annotations\AbstractAnnotation" method="__construct" argument="0" targetInArray="key">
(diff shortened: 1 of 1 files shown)


DerManoMann:
First question that pops into my mind is: who is going to maintain this?

DerManoMann:
It does work, though :)

I just wonder if this is duplicating things. Ideally, I think I'd prefer more docblocks as those are supported by more tools.
Maybe a feature opportunity for the plugin to also support something like this?
```php
   /**
    * @param 'path'|'header'|'query'|'cookie'|null      $in
    */
```

xepozz:
> First question that pops into my mind is: who is going to maintain this?

It does require full-time support. Otherwise, it could be setup one time and after you need just to add missing values or rename classes/methods along with renaming the source. PHPStorm does it automatically, because XML file contains real references to the programming entities.



> It does work, though :)
> 
> I just wonder if this is duplicating things. Ideally, I think I'd prefer more docblocks as those are supported by more tools. Maybe a feature opportunity for the plugin to also support something like this?
> 
> ```
>    /**
>     * @param 'path'|'header'|'query'|'cookie'|null      $in
>     */
> ```

Not many tools support this construction. At first, try to add `(` and `)` around it. Or you may use enums :)
Honestly, it's not goal for the plugin to fix. I hope you may report Jetbrains and they'll fix it. 

Comparing phpdocs vs plugin:
- Plugin may provide additional advantages like item documentation (like PHPStorm has for regexp), icons, etc
- Users may extend vendor plugin config by adding their own custom formats, media types, etc
- PHPDocs for documentation purposes, plugin for interacting with the environment

DerManoMann (approved):
Worth a try; could be interesting.

xepozz:
Aren't there more instructions to add to the config?

I've described only those I use, hope there are some left

PR #1724, opened 2025-02-26
Support `x-enumNames` for `BackedEnum`s

BusterNeece:
Currently, in the OpenAPI spec, there's no standardized way to persist the name of enums when the value of the enum is stored in the `enum` value. There are competing implementations for providing this data in the form of extensions, but one with widespread support is `x-enumNames`, used by NSwag, `swagger-api-typescript`, and SwaggerUI via a plugin.

This PR specifically expands the enum processing to automatically generate the `x-enumNames` values in cases where the enum name isn't used as the value. This ensures that we're able to use this data in other libraries.

I figured I'd offer it as a PR since it may be of utility to more than just me, but I realize that it's using an extension and I can also implement it via directly modifying the pipeline myself. Either way, perhaps it's helpful for folks!

### DIFF
--- a/src/Processors/ExpandEnums.php
+++ b/src/Processors/ExpandEnums.php
@@ -56,6 +56,15 @@ class ExpandEnums
                     return ($useName || !($case instanceof \ReflectionEnumBackedCase)) ? $case->name : $case->getBackingValue();
                 }, $re->getCases());
 
+                if (!$useName) {
+                    $schemaX = Generator::isDefault($schema->x) ? [] : $schema->x;
+                    $schemaX['enumNames'] = array_map(function ($case) {
+                        return $case->name;
+                    }, $re->getCases());
+
+                    $schema->x = $schemaX;
+                }
+
                 $schema->type = $useName ? 'string' : $enumType;
 
                 $this->mapNativeType($schema, $schemaType);
--- a/src/Processors/ExpandEnums.php
+++ b/src/Processors/ExpandEnums.php
@@ -22,7 +22,7 @@ class ExpandEnums
 
     protected ?string $enumNames;
 
-    public function __construct(string $enumNames = null)
+    public function __construct(?string $enumNames = null)
     {
         $this->enumNames = $enumNames;
     }
@@ -41,7 +41,7 @@ class ExpandEnums
      * x-enumNames:
      *   - NAME1
      *   - NAME2
-     * ```
+     * ```.
      */
     public function setEnumNames(?string $enumNames = null): void
     {
--- a/src/Processors/ExpandEnums.php
+++ b/src/Processors/ExpandEnums.php
@@ -20,22 +20,32 @@ class ExpandEnums
 {
     use Concerns\TypesTrait;
 
-    /** @var bool Whether to add the extension `x-enumNames` to backed enums (default false). */
-    protected bool $addXEnumNames;
+    protected ?string $enumNames;
 
-    public function __construct(bool $addXEnumNames = false)
+    public function __construct(string $enumNames = null)
     {
-        $this->addXEnumNames = $addXEnumNames;
+        $this->enumNames = $enumNames;
     }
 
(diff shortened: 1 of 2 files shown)


DerManoMann:
I like it. Only thing I think should be changed is to make this optional.

Also, this page suggests a variety of different `x-` keys, so maybe allowing to specify it (at the same time) would make it even more useful...
https://stackoverflow.com/questions/66465888/how-to-define-enum-mapping-in-openapi

```php
    protected ?string $enumNames;

    public function __construct(?string $enumNames = null)
    {
        $this->enumNames = $enumNames;
    }

    public function getEnumNames(): ?string
    {
        return $this->enumNames;
    }
```

And then the check would be 
```php
                if (!$useName && $this->enumNames) {
                    // ...
                    $schemaX[$this->enumNames] = array_map(function ($case) {


```

BusterNeece:
I haven't had the time to get back to this as a proper PR, but just as a note, I did go ahead and implement this as a standalone processor just to ensure it worked well for our purposes (and it works perfectly): https://github.com/AzuraCast/AzuraCast/blob/6fed28d138ed9900f7abee41d142ef7ed0407687/backend/src/OpenApi/AddXEnumNames.php

DerManoMann:
@BusterNeece Any update or ETA on this? I feel it is usefult, so happy to help getting it over the line if you do not have spare time to spend.

DerManoMann:
Yeah, sorry, new tests :) That's why having the config option is good/important, so no BC break.

BusterNeece:
@DerManoMann No worries, just updating my branch to start working on it.

BusterNeece:
@DerManoMann So, looking through the code, kinda curious...it doesn't appear there's any example in existing code of one of the processors using a config provided to the Generator as part of its own `config` object.

I could make it an argument of the constructor in the `ExpandEnums` processor, but modifying that would require removing/re-adding it to the stack, which is something I don't mind doing on my end but might be inconvenient for folks.

Thoughts?

BusterNeece:
@DerManoMann Update: Disregard previous post...I see how configuration options are specified via other processors and have mirrored that.

DerManoMann:
Yep that does the job. However ;)

- With processor options, the (admittedly unwritten) contract is that:
  * the default is set on the contructor parameter
  * any phpdoc comments are expected to be made on the corresponding `setXXX` method.
  * Defaults do not need to be documented as they are pulled from the constructor.
  if you look for `setXXX()` methods in the processors folder you'll see they all do it the same way.
- Not sure if you missed this, but in my code sample I made the config option not a `bool`, but a string so the user can pick the actual name. I think I would prefer that as it allows to cater for different downstream tools. The docblock could perhaps suggest one or two of the more common ones (like `enumNames`).
However, I would expect anyone using this feature would know what name they want...

BusterNeece:
@DerManoMann Sorry for the confusion. I did indeed miss that you had specified a complete example above.

My tooling's a little out of whack as I normally use Dockerized containers for all of my PHP dev work. We should be good now, though!

src/Processors/ExpandEnums.php
```diff
@@ -20,22 +20,32 @@ class ExpandEnums
 {
     use Concerns\TypesTrait;
 
-    /** @var bool Whether to add the extension `x-enumNames` to backed enums (default false). */
-    protected bool $addXEnumNames;
+    protected ?string $enumNames;
 
-    public function __construct(bool $addXEnumNames = false)
+    public function __construct(string $enumNames = null)
```
DerManoMann:
```suggestion
    public function __construct(?string $enumNames = null)
```

DerManoMann (changes requested):
Nice, one more nit and it is a wrap!

src/Processors/ExpandEnums.php
BusterNeece: Already in after the CS fix!

PR #1744, opened 2025-03-20
Fix duplicate attachables

smallsung:
Fix duplicate `attachables` and format `attachables` assignment.

Bug display:
https://github.com/zircote/swagger-php/blob/18457fa71f753cfd4a2b21916baf329864fdfaa6/src/Attributes/Schema.php#L110-L111


DerManoMann:
Thanks @smallsung

PR #1748, opened 2025-03-27
Introduce `Generator` settings

### LINKED ISSUE #1747: Ignore other annotations
For other annotations outside of openapi, I have another program analysis. I don't want to construct these annotations.

DerManoMann:
Generator settings are following the same pattern as processor config and are stored under the key `'generator'`.

`Generator::getSetting()` provides a shortcut to access generator settings.

Fixes #1747 by allowing to set `Generator::setConfig(['generator' => ['ignoreOtherAttributes' => true]]);`

### DIFF
--- a/bin/openapi
+++ b/bin/openapi
@@ -218,6 +218,7 @@ foreach ($options["processor"] as $processor) {
 }
 
 $analyser = new ReflectionAnalyser([new DocBlockAnnotationFactory(), new AttributeAnnotationFactory()]);
+$analyser->setGenerator($generator);
 
 $openapi = $generator
     ->setVersion($options['version'])
--- a/src/Analysers/AnalyserInterface.php
+++ b/src/Analysers/AnalyserInterface.php
@@ -8,11 +8,8 @@ namespace OpenApi\Analysers;
 
 use OpenApi\Analysis;
 use OpenApi\Context;
-use OpenApi\Generator;
 
-interface AnalyserInterface
+interface AnalyserInterface extends GeneratorAwareInterface
 {
-    public function setGenerator(Generator $generator): void;
-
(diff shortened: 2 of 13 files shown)


smallsung:
This is what I need, I think it can be designed to be more flexible.

Like:
```php
$attributeArgs = ['IgnoreAnnotationClass1','IgnoreAnnotationClass2'];

foreach ($reflector->getAttributes() as $attribute) {
    if (in_array($attribute::class, $attributeArgs, true)) {
    	continue;
    }
}
```
or
```php
interface IgnoreAttributes{}

foreach ($reflector->getAttributes() as $attribute) {
    if ($attribute instanceof IgnoreAttributes) {
    	continue;
    }
}
```
or to be more precise
```php
interface IgnoreAttributes{
	public function ignoreOAAttributes():bool;
}

foreach ($reflector->getAttributes() as $attribute) {
    if ($attribute instanceof IgnoreAttributes && $attribute->ignoreOAAttributes()) {
    	continue;
    }
}
```

I prefer the second or third one.

smallsung:
If you agree and don't have extra time, I can pr.

DerManoMann:
I think I am still not quite sure what the actual requirements are.

Excluding any attributes that are not OpenApi attributes should be all yours - unless you extend from `AbstractAnnotation`. But then I would expect that the library should process them...??

Could you please define what the actual criteria is to exclude attributes (without any new interface/logic, etc.)?

DerManoMann:
Having second thoughts about the way the settings are done, but other than that I think this is good.

smallsung:
This pr meets my needs perfectly.

DerManoMann:
Closing in favour of #1756

PR #1761, opened 2025-04-24
[v6] - PHP 8.1

### LINKED ISSUE #1669: [v6] Bump PHP deps to 8.1
Just a place holder for now and to remind people to upgrade. Right now this would affecgt about 15% of installs, so maybe next year :)

DerManoMann:
Fixes #1669 

- [ ] Update general docs
- [ ] Flesh out migration guide
- [ ] Refactor `Util` helper
- [ ] Remove deprecated elements
- [ ] Deprecate `Generator::scan()`
- [ ] Add phpdoc parser / type handling refactor
- [ ] ...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: true
       matrix:
         operating-system: [ ubuntu-latest ]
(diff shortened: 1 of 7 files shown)


Kiina:
So is the type-info-extras not landing in v6? Since it requires 8.2 while the min version is bumped to 8.1.

PR #1764, opened 2025-04-25
FIX tests for windows

srsbiz:
Summary of changes for tests to work in windows environments:

- normalized paths where possible for assertions
- use DIRECTORY_SEPARATOR inside paths where normalization is not possible
- prepend `php` to command line tests
- use `NUL` instead of `/dev/null` for error output in command line tests

### DIFF
--- a/tests/CommandlineTest.php
+++ b/tests/CommandlineTest.php
@@ -12,7 +12,7 @@ class CommandlineTest extends OpenApiTestCase
 {
     use UsesExamples;
 
-    private function getCommandToExecute(string $cmd, bool $appendDevNull = true): string
+    private function getCommandToExecute(string $cmd, ?string $devNullRedir = null): string
     {
         if (PHP_OS_FAMILY === 'Windows') {
             $cmd = 'php ' . $cmd;
@@ -20,8 +20,8 @@ class CommandlineTest extends OpenApiTestCase
         } else {
             $devNull = '/dev/null';
         }
-        if ($appendDevNull) {
-            $cmd .= ' ' . $devNull;
+        if ($devNullRedir) {
+            $cmd .= " $devNullRedir $devNull";
         }
 
         return $cmd;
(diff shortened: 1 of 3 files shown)


tests/CommandlineTest.php
```diff
@@ -12,22 +12,37 @@ class CommandlineTest extends OpenApiTestCase
 {
     use UsesExamples;
 
+    private function getCommandToExecute(string $cmd, bool $appendDevNull = true): string
```
DerManoMann, line 15: I think it would simplify things even more if we optionally passed in the redirect and make 2nd paramter `?string $devNullRedir = null`.

Then we can call it like 
```php
   $this->getCommandToExecute('xxx', '2>');
```

and the function would do:
```php
   if ($devNullRedir) {
       $cmd .= " $devNullRedir $devNull";
   }
```

tests/UtilTest.php
```diff
@@ -9,6 +9,7 @@
 use OpenApi\Annotations as OA;
 use OpenApi\Tests\Concerns\UsesExamples;
 use OpenApi\Util;
+use Symfony\Component\Filesystem\Path;
```
DerManoMann, line 12: This makes `symfony/filesystem` a direct dependency, so it should be added to `composer.json`

DerManoMann (changes requested):
Looking ok to me. Happy to merge if you at least update composer.json

tests/CommandlineTest.php
srsbiz: Done

tests/UtilTest.php
srsbiz: After closer look, that would require php >= 8.0.2, so I have dropped this dependency and used simple lambda function instead.

DerManoMann (approved):
I think there is still an opportunity to use short functions but I feel this will do for now.
`v6` will bring some changes to tbe `Util` class, maybe that is the time to do some more cleanup here.

DerManoMann:
Thanks @srsbiz

PR #1767, opened 2025-05-07
Additions to AttributeAnnotationFactory

Onekone:
While developing my own set of helper annotations, noticed that context object lacks information to which method argument attribute is attached to (case is to use it in processing* to implicitly fill out attributes), so this changes that, and adds reflection instances for good measure

Also makes it so all OA/Schemas in arguments are evaluated, and not just OA/Properties

----
\* And in a hacky way, during validation


igs-raiderler:
Excellent contribution to open source! The community will benefit greatly from this.

DerManoMann:
Could you give an example of how/why there might be a `OA\Schema` on a method parameter?

Onekone:
Not exactly exact OA/Schema, but for example, a custom attribute extending OA/Schema, that targets method parameters and gets filled from respective docblock `@param` annotation containing array/object shape. 

Closest practical example would be a Laravel Eloquent model that has JSON field (which would be `array|object|null`), and a setter method. Defining input argument typehint as an object shape (ex. `object{ items: string[], count: non-negative-int, type: "box"|"bag"}`), and giving it a such custom schema attribute can give you a schema that can be reused in a compact way 

I am brainstorming it myself currently (see onekone/lore repo, but it's deeply wip), so forgive me if I'm not convincing enough

DerManoMann:
@Onekone have you looked at the latest changes around type resolving and in particular #1829 ?
I think once that is merged I will close this PR as I think just tracking the single `Context::resolver` gives us everything you would need.

DerManoMann:
Closing - stale and superseded by the new `Context::reflector` property.

Onekone:
Sorry for long response, was a bit busy on personal and work side of things

Yeah, it's fine, current implementation also works great

PR #1775, opened 2025-07-10
Fix parameter type

krissss:
For Fix nativeTypeMap not execd in PathParameter only used with no docblock

image

image

### DIFF
--- a/tests/Fixtures/RequestUsingAttribute.php
+++ b/tests/Fixtures/RequestUsingAttribute.php
@@ -4,8 +4,6 @@
  * @license Apache 2.0
  */
 
-// NOTE: this file uses "\r\n" linebreaks on purpose
-
 namespace OpenApi\Tests\Fixtures;
 
 use OpenApi\Attributes as OAT;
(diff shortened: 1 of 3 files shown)


deluxetom (approved):

DerManoMann:
Looks reasonable. Could I ask for a test to be added (that fails on current master) to cover this change? Changing an existing test is fine too.

krissss:
> Looks reasonable. Could I ask for a test to be added (that fails on current master) to cover this change? Changing an existing test is fine too.

added

tests/Fixtures/RequestUsingAttribute.php
```diff
@@ -0,0 +1,23 @@
+<?php declare(strict_types=1);
+
+/**
+ * @license Apache 2.0
+ */
+
+// NOTE: this file uses "\r\n" linebreaks on purpose
```
DerManoMann, line 7: I think this is just a cut/paste leftover that can be removed?

DerManoMann (changes requested):
One more nit then I think it's done. Nice work.

DerManoMann (approved):
Thanks @krissss

PR #1776, opened 2025-07-17
Support use x-* in ref

krissss:
```php
@OA\Schema(
     schema="String",
     @OA\Property(property="value", type="string"),
     x={
        "custom-key": @OA\Schema(
            @OA\Property(property="value", type="string"),
        )
    }
)
```

The above annotation will generate `x-custom-key` in the OpenAPI document.

However, we need to get ref of `x-custom-key` with `$openapi->ref('#/components/schemas/String/x/custom-key')` (note that here we use `x/custom-key` instead of `x-custom-key`).

This is not conducive to writing ref in the code ourselves. For example,

```php
@OA\Post(
    path="/api/~/endpoint",
    @OA\Response(
        response="default",
        description="A response",
        @OA\JsonContent(ref="#/components/schemas/String/x/custom-key")
    )
)
```

This will generate the following incorrect documentation when using `$openapi->toYaml()`(use `x/custom-key`)

```yaml
...
            application/json:
              schema:
                $ref: '#/components/schemas/String/x/custom-key'
...
```

Therefore, we have to write it in the following way (use `x-custom-key`)

```php
@OA\Post(
    path="/api/~/endpoint",
    @OA\Response(
        response="default",
        description="A response",
        @OA\JsonContent(ref="#/components/schemas/String/x-custom-key")
    )
)
```

However, if it is written directly in the above way, the validation will fail.

This PR is to fix this issue to support the use of ref `x-*`.

### DIFF
--- a/src/Annotations/OpenApi.php
+++ b/src/Annotations/OpenApi.php
@@ -219,11 +219,21 @@ class OpenApi extends AbstractAnnotation
         $unresolved = $slash === false ? $resolved . $subpath : $resolved . $subpath . '/';
 
         if (is_object($container)) {
-            if (property_exists($container, $property) === false) {
+            // support use x-* in ref
+            $xKey = strpos($property, 'x-') === 0 ? substr($property, 2) : null;
+            if ($xKey) {
+                if (!is_array($container->x) || !array_key_exists($xKey, $container->x)) {
+                    $xKey = null;
+                }
+            }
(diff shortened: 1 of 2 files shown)


DerManoMann:
Interesting - I didn't even know you can nest complex structures in `x`. Somehow I always assumed it was a flat map...

DerManoMann (approved):
Nice!

DerManoMann:
Thanks @krissss

PR #1785, opened 2025-08-11
Favour promoted property annotation nullable over native

### LINKED ISSUE #1783: Setting 'nullable: false' doesn't affect when variable has a type ?string (?integer, etc)
### Problem
I have a code
```
#[OA\Property(nullable: false)]
public ?string $signed_at,
```

in generated json it's described as
```
"signed_at": {
   "type": "string",
   "nullable": true
},
```

Expected behaviour should be
```
"signed_at": {
   "type": "string",
   "nullable": false
},
```

### Why I think it's an issue
In all 3 cases
```
#[OA\Property(nullable: true)]
public ?string $signed_at,

#[OA\Property(nullable: false)]
public ?string $signed_at,

#[OA\Property()]
public ?string $signed_at,
```
result is the same
```
"signed_at": {
   "type": "string",
   "nullable": true
},
```

BUT
if I set variable as 
```
public string $signed_at,
```
then in all three cases
```
#[OA\Property(nullable: true)]
public string $signed_at,

#[OA\Property(nullable: false)]
public string $signed_at,

#[OA\Property()]
public string $signed_at,
```
I retrieve totally different result

```
"signed_at": {
   "type": "string",
   "nullable": true
},

"signed_at": {
   "type": "string",
   "nullable": false
},

"signed_at": {
   "type": "string",
},
```

I would say that the behaviour is not consistent. In case of `string`, we fully trust a developer and nullable attribute description, but in case of `?string` we don't trust a developer and nullable attribute description.

### Use case
In our backend, we remove all nullable variables from response DTOs and never send nulls. (but internally in our DTO it's nullable before sending.) From the frontend point of view, our backend API either returns a variable with a value or do not return the variable at all. Case "we return a variable with null" is not possible. However, I'm unable to create a correct Swagger documentation for the API due to this issue.

DerManoMann:
Fixes #1783 (part II)

### DIFF
--- a/src/Analysers/AttributeAnnotationFactory.php
+++ b/src/Analysers/AttributeAnnotationFactory.php
@@ -85,7 +85,9 @@ class AttributeAnnotationFactory implements AnnotationFactoryInterface
                                 if (Generator::isDefault($instance->type)) {
                                     $instance->type = $type;
                                 }
-                                $instance->nullable = $nullable ?: Generator::UNDEFINED;
+                                if (Generator::isDefault($instance->nullable)) {
+                                    $instance->nullable = $nullable ?: Generator::UNDEFINED;
+                                }
 
                                 if ($rp->isPromoted()) {
                                     // ensure each property has its own context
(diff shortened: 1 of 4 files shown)


dima-stazzle:
@DerManoMann Can you please create a new release with that?

DerManoMann:
Sure - ideally you would test first ;) but I think it is pretty safe.

dima-stazzle:
It works like a charm. Thanks a lot!

PR #1801, opened 2025-09-01
Add generic type resolvers

### LINKED ISSUE #1310: Refactor type handling
Needs to include:
- simple types
-  type mapping PHP -> OAS
- ReflectionUnionType
- ReflectionIntersectionType

### LINKED ISSUE #1318: Different state of real class properties and schema properties: nullable
Hello. Thank you for your package, it has saved some my time due to existent documentation in classes. 

But I found a feature or a bug) when the real properties of a class are different with schema properties. 

**Steps to reproduce**

I have the next code.

```php
    /**
     * Some description.
     *
     * @OA\Property(nullable=false, format="number", example="0.00")
     */
    public ?string $value = null;
```

It describes the value that has the default value `null`, but this value is never returned in responses and there is always non-null values, like *0.00*.

I expected the follow description in *yaml*. There is `nullable: false` or without `nullable` due to I redefine this property.

```yml
      properties:
        annual:
          description: 'Some description.'
          type: string
          nullable: false # here
          format: number
          example: '0.00'
```

or

```yml
      properties:
        annual:
          description: 'Some description.'
          type: string # there isn't nullable here
          format: number
          example: '0.00'
```

But I got the follow result.

```yml
      properties:
        annual:
          description: 'Some description.'
          type: string
          nullable: true # here
          format: number
          example: '0.00'
```

To fix it I've redefined the value type: `type="string"`.

```php
    /**
     * Some description.
     *
     * @OA\Property(type="string", format="number", example="0.00")
     */
    public ?string $value = null;
```

Now I got this result.

```yml
      properties:
        annual:
          description: 'Some description.'
          type: string # there isn't nullable here, because I defined another type.
          format: number
          example: '0.00'
```

I would like to know, it is a bug or a feature? Perhabs the tool should override the detected state of the property. 

Thank you!

### LINKED ISSUE #1354: If a type of a defined property is nullable, nullable attribute should be added to the generated Doc
Probably this is more like a feature request.

For example when the property is defined like this:
```php
#[OA\Property(description: 'Name')]
public readonly ?int $name;
```

The following should be generated into the documentation:
```
"name": {
    "description": "Name",
    "type": "integer",
    "nullable": true
}
```

### LINKED ISSUE #1676: Array shapes support
I've been trying to get my custom processor to get a good swagger.json for arrays. I've tried to use the PhpDocParser but noticed it gets a bit complex after a while. Then I looked at how doctypes are handled by and saw the [extractVarTypeAndDescription](https://github.com/zircote/swagger-php/blob/7df10e8ec47db07c031db317a25bef962b4e5de1/src/Processors/Concerns/DocblockTrait.php#L177) method. It uses some regex that won't really work with [array shapes](https://psalm.dev/docs/annotating_code/type_syntax/array_types/#array-shapes).

Let's take a simple one:

`/* @var array{id: int, name: string} */`

Calling the extract on it returns:

```php
array(2) {
  ["type"]=>
  string(9) "array{id:"
  ["description"]=>
  string(18) "int, name: string}"
}
```

Which results in:

```yml
properties:
  name:
    description: 'int, name: string}'
```

Which isn't the desired result.

### LINKED ISSUE #1678: [v6?] add phpdoc parser
https://github.com/phpstan/phpdoc-parser

DerManoMann:
Fixes #1678, fixes #1676, fixes #1354, fixes #1318, fixes #1310

Adds type resolvers to normalize and improve type handling.

- [x] Add type resolver interface and `type-info` based resolver
- [x] Add legacy resolver  
- [x] Use to resolve property types
- [x] Use to resolve other types ...?
- [x] Define/document strategy of how types are resolved

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -310,17 +310,19 @@ class Generator
 
     public function getTypeResolver(): TypeResolverInterface
     {
-        return $this->typeResolver ?:
-            (
-                class_exists(StringTypeResolver::class)
-                ? new TypeInfoTypeResolver()
-                : new LegacyTypeResolver()
-            );
+        $this->typeResolver ??=                class_exists(StringTypeResolver::class)
+                    ? new TypeInfoTypeResolver()
+                    : new LegacyTypeResolver();
+        ;
+
+        return $this->typeResolver;
     }
 
     public function getLogger(): ?LoggerInterface
     {
-        return $this->logger ?: new DefaultLogger();
+        $this->logger ??= new DefaultLogger();
+
+        return $this->logger;
     }
 
     public function getVersion(): ?string


DerManoMann:
**Type resolving strategy:**
* Use OpenApi schema/type
* Use docblock type-hints
* Use native (reflection) type-hints

**Applicable to:**
* `nullable`
* `items`
* `ref`
* `type`
* `minimum`/`maximum`
* `const`
* `enum`


BC break: Unknown/invalid types (not native, not class/interface/enum) will be dropped. This is a change from the current behaviour to keep both legacy and type-info resolver in line.

DerManoMann:
I think I'll leave it here. There will be follow ups as there is more scope to use the new resolvers in other places...

DerManoMann:
That being said, this is pretty close to being merged :)

momala454:
Since this pull request, properties without type now get a type of "integer" instead.
My property have a "oneOf" (with 2 classes as type ) to describe the real type, so it isn't supposed to have integer

DerManoMann:
@momala454 could you please open a new issue and paste an example? Thanks.

Kiina:
Just a quick question, whats the recommended way of turning my array shape into something usable for the swagger file?

For non arrays I used to do something like this:

```php
$this->mapNativeType($schema, strtolower($type));
$newProperties[] = new OpenApi\Attributes\Property(
    property: $constantValue,
    type: $schema->type,
    format: $schema->format,
    pattern: !empty($regexPattern) ? $regexPattern : Generator::UNDEFINED,
    nullable: $returnType->allowsNull()
);
```


Now i have a reflection on a function that returns an array shape, when I do something like this:

```php
$test = (new OpenApi\Type\TypeInfoTypeResolver)->getDocblockTypeDetails($methodReflection);
var_dump($test);
```

I get:

```php
class stdClass#38609 (6) {
  public $explicitType =>
  string(47) "array{'id': int, 'offset': int, 'priority': int}"
  public $explicitDetails =>
  NULL
  public $types =>
  array(1) {
    [0] =>
    string(47) "array{'id': int, 'offset: int, 'priority': int}"
  }
  public $name =>
  string(7) "getPriority"
  public $nullable =>
  bool(false)
  public $isArray =>
  bool(true)
}
```

since the signature is:

```php
/**
 * @return array{
 *     id: int,
 *     priority: int,
 *     offset: int
 * }[]
 */
```

But I'm a bit lost on how i would get a property out of it to add it to my schema

DerManoMann:
@Kiina the legacy code is not really good enough to handle multi-line docblock types. Its mostly a slightly improved version of what was there before, so you are out of luck if you are stuck with it I am afraid.

I'll have to re-think the type resolvers a bit, I think, in particular to actually properly support these cases with the new (8.2+) TypeInfoTypeResolver.

PR #1802, opened 2025-09-01
docs: mention Scalar as an alternative to Swagger UI

hanspagel:
big PHP fan here, I just discovered swagger-php. I spend the last years to build an alternative to Swagger UI that some people like, mostly for the more modern look and the fully featured API client that’s embedded. it’s called [Scalar](https://github.com/scalar/scalar).

I just thought it would be cool to offer it as an alternative to Swagger UI, what do you think?

I added it in two places. If you think that's too much, I can also just put it in one place, or whatever you think is good.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,6 @@
 ---
 layout: home
-hero: 
+hero:
   name: Swagger-PHP
   tagline: Generate OpenAPI documentation for your RESTful API.
   actions:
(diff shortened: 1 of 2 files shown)


DerManoMann:
Never heard of that before - nice! Let's go with it - if I mention one I can also mention two I suppose. If there are more to come we can always rething.

PR #1829, opened 2025-10-29
Type resolver refactor

### LINKED ISSUE #1340: The 'items' field for Array property's isn't populated by the DocBlock
The following constructor:

```
     /**
     * @param DateTimeImmutable[] $times
     * @param string[] $alerts
     */
    public function __construct(
        #[OA\Property]
        public readonly array $times,
        #[OA\Property]
        public readonly array $alerts,
    )
````

Will throw the error: `@OA\Items() is required when @OA\Property() has type "array"`

The fix is to manually populate the info, this works:

     /**
     * @param TeamRetainedSalary[] $retainedSalaryTeams
     * @param string[] $alerts
     */
    public function __construct(
        #[OA\Property(items: new OA\Items(ref: "#/components/schemas/DateTimeImmutable", type: "object"))]
        public readonly array $times,
        #[OA\Property(items: new OA\Items(type: "string"))]
        public readonly array $alerts,
    )
````

Is this a bug? Should the generator automatically populate the items info from the DocBlock generics?

### LINKED ISSUE #1456: Unable to reuse RequestBody definitions via FQCN references
Version of `swagger-php`: `4.7.10`
Version of PHP: `8.2.7`

Hello. I am uncertain as to whether this is a bug or a feature that is not present.

## Desired Behaviour

I would like to reference top-level `OA\RequestBody` definitions from operation definitions using FQCNs (fully qualified class names). The [OpenApi 3.0 spec (§ 4.7.10.1)](https://spec.openapis.org/oas/v3.0.0#operationRequestBody) as well as [Swagger docs (§ Reusable Bodies)](https://swagger.io/docs/specification/describing-request-body) seem to indicate this as a possibility.

#### Attribute Declarations

```php
#[OA\Post(
    path: '/departments',
    requestBody: new OA\RequestBody(ref: StoreDepartmentRequest::class),
)]
public function store(StoreDepartmentRequest $request): DepartmentResource {}
```

```php
#[OA\RequestBody(
    request: 'StoreDepartmentRequest',
    content: new OA\JsonContent(
        properties: [
            new OA\Property(property: 'name', type: 'string')
        ]
    )
)]
class StoreDepartmentRequest extends FormRequest {}
```

#### Generated OpenAPI YAML

If I understand the OpenAPI spec correctly, the desired OpenAPI document to generate would have the following fragments:
```yaml
paths:
  /departments:
    post:
      operationId: fa715e92b3f13a22e61601b13cb6921c
      requestBody:
        $ref: '#/components/requestBodies/StoreDepartmentRequest'
```

```yaml
components:
  requestBodies:
    StoreDepartmentRequest:
      content:
        application/json:
          schema:
            properties:
              name:
                type: string
            type: object
```

## Actual Behaviour

Inspecting the actual generated OpenAPI YAML, the value under `paths.?.post.requestBody.$ref` does not seem to be properly processed: instead of a `#/components/requestBodies/...` URI, the value remains the same as the initial FQCN.

```yaml
paths:
  /departments:
    post:
      operationId: fa715e92b3f13a22e61601b13cb6921c
      requestBody:
        $ref: App\Http\Requests\StoreDepartmentRequest

```

## Seemingly pertinent parts of codebase 
While I am not totally familiar with the `swagger-php` codebase, I noticed that the following methods seem responsible for the current behaviour:
* [`\OpenApi\Processors\AugmentRefs::resolveFQCNRefs`](https://github.com/zircote/swagger-php/blob/7e583e122db46602bf1640a9cb3473f2b83d1b8a/src/Processors/AugmentRefs.php#L58-L71)
* [`\OpenApi\Analysis::getSchemaForSource`](https://github.com/zircote/swagger-php/blob/7e583e122db46602bf1640a9cb3473f2b83d1b8a/src/Analysis.php#L326-L344)

### LINKED ISSUE #1819: Type handling refactor II
Follow up for #1801 to look at other places where the resolvers could be used.

Considering that the type resolver code is pretty fresh I think this will be done as a minor release bump. Current interface methods have already been marked `deprecated`

### LINKED ISSUE #1826: radebatz/type-info-extras add an invalid type
> The following
> ```php
> #[\OpenApi\Attributes\Schema()]
> class Test
> {
>     /**
>      * @param string|list<string> $value
>      */
>     public function __construct(
>         #[\OpenApi\Attributes\Property(example: 'My value')]
>         public string|array $value,
>     ) {
>     }
> }
> ```
> 
> Will generate this without type info extras
> ```json
> "Test": {
>                 "properties": {
>                     "value": {
>                         "example": "My value"
>                     }
>                 },
>                 "type": "object"
>             }
> ```
> 
> But it generates this with type info extras
> ```json
> "Test": {
>                 "properties": {
>                     "value": {
>                         "type": "array",
>                         "items": [],
>                         "example": "My value"
>                     }
>                 },
>                 "type": "object"
>             }
> ```
> 
> I don't know what I should expect, but only "array" as type isn't really correct
> 
> Should I create an issue ? 

 _Originally posted by @momala454 in [#1822](https://github.com/zircote/swagger-php/issues/1822#issuecomment-3414338950)_

DerManoMann:
Fixes #1819, fixes #1826, fixes #1340, fixes #1456

Drops support for:
* `Context::type`
* `Context:nullable`

as those are now fully controlled by the type resolver.

The reason for some of the BC breaking changes is that this PR switches to using the same rules about type resolving for all annotation/attribute properties and removes some exceptional logic.

### DIFF
--- a/tests/ScratchTest.php
+++ b/tests/ScratchTest.php
@@ -49,7 +49,7 @@ class ScratchTest extends OpenApiTestCase
      *
      * @dataProvider scratchTestProvider
      *
-     * @requires     PHP 8.1
+     * @requires     PHP 8.2
      */
     public function testScratch(string $scratch, string $spec, string $version, array $expectedLogs): void
     {


DerManoMann:
Slowly starting to feel good about these changes :)

PR #1835, opened 2025-11-10
Add `Encoding` annotation to manage `MediaType` encodings

### LINKED ISSUE #1062: Will support MediaType Encode Object?
https://github.com/zircote/swagger-php/blob/8f53ded4c90ec7cccb18333c1d7d8a4c16f0c3c8/src/Annotations/MediaType.php#L61

DerManoMann:
Fixes #1062

Breaking BC: Changes the `MediaType::encoding` type from `array<string,mixed>` to `Encoding[]`

Does what it says on the box.

In addition, this also now allows to set the new `Encoding` on `Schema` properties (`Property`) as a shortcut, as it keeps the two together and means that `Encoding::property` can be inferred and doesn't need to be set twice (and kept in sync)

Example: https://github.com/DerManoMann/swagger-php/blob/encoding/tests/Fixtures/Scratch/Encoding.php#L112-L114

### DIFF
--- a/src/Annotations/Header.php
+++ b/src/Annotations/Header.php
@@ -95,6 +95,7 @@ class Header extends AbstractAnnotation
      * @inheritdoc
      */
     public static $_parents = [
+        Encoding::class,
         Components::class,
         Response::class,
     ];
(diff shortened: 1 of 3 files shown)


melroy89:
This is a breaking change.. 😢  https://github.com/zircote/swagger-php/issues/1841

Yet it was not a major version upgrade, and just rolled-out as a minor version dump..

PR #1838, opened 2025-11-12
Add encoding property

### LINKED ISSUE #1837: Issues with v5.7.0 in combination with nelmio-api-docs bundle
We are running the nelmio-api-docs bundle and with the update of this package to 5.7.0 our CI started to fail with the message 

``` [ErrorException]                                                                                                                                                                                                              
  User Warning: Ignoring unexpected property "encoding" for @OA\Schema(), expecting "x", "ref", "schema", "title", "description", "maxProperties", "minProperties", "required", "properties", "type", "format", "items", "coll  
  ectionFormat", "default", "maximum", "exclusiveMaximum", "minimum", "exclusiveMinimum", "maxLength", "minLength", "pattern", "maxItems", "minItems", "uniqueItems", "enum", "multipleOf", "discriminator", "readOnly", "writ  
  eOnly", "xml", "externalDocs", "example", "examples", "nullable", "deprecated", "allOf", "anyOf", "oneOf", "not", "additionalProperties", "additionalItems", "contains", "patternProperties", "dependencies", "propertyNames  
  ", "const", "contentEncoding", "contentMediaType"    
```

It looks like in #1835 is a 

```
public $encoding = Generator::UNDEFINED;
```

missing somewhere behind https://github.com/zircote/swagger-php/blob/39fcd46e79c2f3cfbf56cf5a92a86108c8eed401/src/Annotations/Schema.php#L445

heiglandreas:
Otherwise toolage fails when trying to add data with an encoding attribute that is not known in the properties here

This fixes #1837

### DIFF
--- a/src/Annotations/Schema.php
+++ b/src/Annotations/Schema.php
@@ -445,6 +445,11 @@ class Schema extends AbstractAnnotation
     public $contentMediaType = Generator::UNDEFINED;
 
     /**
+     * @var string
+     */
+    public $encoding = Generator::UNDEFINED;
+
+    /**
      * @inheritdoc
      */
     public static $_types = [


DerManoMann:
So, sounds like this is fixed in `NelmioApiDocBundle`, so closing this one.

PR #1844, opened 2025-11-15
Handle Mixed union type incl. `array` in `LegacyTypeResolver`

### LINKED ISSUE #1826: radebatz/type-info-extras add an invalid type

DerManoMann:
Fixes #1826 

Fixes the specific issue, but still leaves room for improvement to deal with non simple type unions.

### DIFF
--- a/src/Type/LegacyTypeResolver.php
+++ b/src/Type/LegacyTypeResolver.php
@@ -202,6 +202,11 @@ class LegacyTypeResolver extends AbstractTypeResolver
         $nullable = in_array('null', explode('|', strtolower($type))) || str_contains($type, '?');
         $isArray = str_contains($type, '[]') || str_contains($type, 'array');
         $type = str_replace(['|null', 'null|', '?', 'null', '[]'], '', $type);
+        $isUnion = count(explode('|', $type)) > 1;
+        if ($isUnion && $isArray) {
+            $type = '';
+            $isArray = false;
+        }
 
         // typed array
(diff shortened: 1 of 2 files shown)


DerManoMann:
Sigh. According to the tests, this is handled the same by both resolvers: https://github.com/zircote/swagger-php/blob/master/tests/Fixtures/Scratch/MultiTypeProperty.php

Not much more I can do without a reproducer - happy to keep fixing as long as I have a target and something to test against.
Sorry for things being so tricky :/

PR #1846, opened 2025-11-16
Treat empty string `''` same as undefined for all string based properties

DerManoMann:
Related to #1668

### DIFF
--- a/docs/examples/specs/petstore/annotations/Models/Pet.php
+++ b/docs/examples/specs/petstore/annotations/Models/Pet.php
@@ -25,9 +25,11 @@ use OpenApi\Annotations as OA;
 class Pet
 {
     /**
+     * The description.
+     *
      * @OA\Property(
      *     format="int64",
-     *     description="ID",
+     *     description="",
      *     title="ID",
      * )
      *
--- a/docs/examples/specs/petstore/attributes/Models/Pet.php
+++ b/docs/examples/specs/petstore/attributes/Models/Pet.php
@@ -26,8 +26,11 @@ use OpenApi\Attributes as OAT;
 )]
 class Pet
 {
+    /**
+     * The description.
+     */
     #[OAT\Property(
-        description: 'ID',
+        description: '',
         title: 'ID',
         format: 'int64'
     )]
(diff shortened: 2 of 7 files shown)


bfanger:
imo: If you document it inside PHP as empty string, the swagger-php output should be an empty string.

DerManoMann:
It's a tricky one IMO. The attribute should be the authority, so explicitly passing in `null` should override the docblock.

The original issue was explicitly about hiding the docblock description - which works for annotations. Maybe it's time to give up feature parity and move on and require an empty string in the attribute, so it can be distinguished from the default of `null` which **does** allow fallback to the docblock.

bfanger:
I agree that the attribute should be the authority _and_ that we need a way to say "don't use the text from the docblock"

But the change in AbstractAnnotation.php affects all properties, not just the description.

When defining
```
#[OAT\Schema(
  schema: "emptystring",
  default: '',
)
```
I'd like the `"default": "",` to be in the generated json output.
 
Wouldn't updating the description in the attribute classes to:
```php
?string $description = Generator::UNDEFINED,
```
solve the issue?

DerManoMann:
I know that the PR is not that great. Also, just using `Generator::UNDEFINED` would work in this case, but I feel either it should be used everywhere as default (instead of the current `null`) or not at all.

I'd love to avoid having to use it as it stops using proper type-hints, whereas something like `?bool` seems reasonable (as opposed to `bool|string`).

Still looking for a solution that would work with `null` as a default value for all constructor arguments but I think PHP just cannot do that.

DerManoMann:
Maybe I am overthinking this. Also this is really just an issue for string values that may fallback to phpdoc strings if not provided, so in reality it is limited to title/description/summary, iirc. Since those are `string` anyway, maybe special handling could be limited to those 3 parameters in attributes.

bfanger:
Agree, we want proper type hints, adding the string type would cause confusion.
Special-casing the title/description/summary would work, i'd prefer using the Generator::UNDEFINED default-value hack and setting the value to `null` to mean omit, but adding a "" case for these properties is also fine. or both.

DerManoMann:
Closing in favour of #1886

PR #1865, opened 2025-12-03
Methods SplObjectStorage::attach(), detach & contains() are deprecated since PHP 8.5.

yblatti:
Methods SplObjectStorage::[attach()](https://www.php.net/manual/en/splobjectstorage.attach.php), [detach()](https://www.php.net/manual/en/splobjectstorage.detach.php) & [contains()](https://www.php.net/manual/en/splobjectstorage.contains.php) are deprecated since PHP 8.5.
Use [offsetSet()](https://www.php.net/manual/en/splobjectstorage.offsetset.php), [offsetUnset()](https://www.php.net/manual/en/splobjectstorage.offsetunset.php) & [offsetExists()](https://www.php.net/manual/en/splobjectstorage.offsetexists.php) instead.

Those are present in PHP 5 >= 5.3.0, PHP 7 and PHP 8, so backward compatibility should be maintained.

### DIFF
--- a/src/Analysis.php
+++ b/src/Analysis.php
@@ -55,7 +55,7 @@ class Analysis
 
     public function addAnnotation(object $annotation, Context $context): void
     {
-        if ($this->annotations->contains($annotation)) {
+        if ($this->annotations->offsetExists($annotation)) {
             return;
         }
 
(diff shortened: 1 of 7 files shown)


DerManoMann:
Thanks @yblatti

PR #1886, opened 2026-01-01
Change attribute `summary`/`description` default value to support `null` as user provided value

DerManoMann:
Alternative solution for #1668


DerManoMann:
@bfanger How about this one then? It was interesting to find out that the behaviour is actually not consistent across all annotations as property/paramameter have their own dedicated processors which meant some extra changes to make `description: null` work the same everywhere...

Issue #1917, opened 2026-01-14
Question: add webhook to OpenApi


qdequippe:
👋🏼 

I am trying to add a webhook (using NelmioApiDocBundle) like this

```php

use Nelmio\ApiDocBundle\Describer\DescriberInterface;
use OpenApi\Annotations\OpenApi;
use OpenApi\Attributes as OA;
use OpenApi\Generator;
use Symfony\Component\DependencyInjection\Attribute\AutoconfigureTag;

#[AutoconfigureTag('nelmio_api_doc.describer')]
final class WebhookDescriber implements DescriberInterface
{
    public function describe(OpenApi $api): void
    {
        if (Generator::UNDEFINED === $api->webhooks) {
            $webhooks = [];
        }

        $webhooks[] = new OA\Webhook(
            webhook: 'test',
            post: new OA\Post(
                requestBody: new OA\RequestBody(
                    description: 'Payload.',
                    required: true,
                    content: new OA\MediaType(
                        mediaType: 'application/json',
                        schema: null,
                    )
                )
            )
        );

        $api->webhooks = $webhooks;
    }
}
```

But an error occured when rendering API doc

```
User Warning: Multiple @OA\Webhook() with the same webhook="test":
// vendor/zircote/swagger-php/src/Annotations/AbstractAnnotation.php:501
```

Webhook entry is duplicate here in OpenApi object https://github.com/zircote/swagger-php/blob/master/src/Processors/MergeIntoOpenApi.php#L76

```
// src/Processors/MergeIntoOpenApi.php

// count($openapi->webhooks) = 1
$openapi->merge($merge, true);
// count($openapi->webhooks) = 2 with duplicate
```
Any idea how to add a custom webhook to api doc?

DerManoMann:
To be perfectly honest, the merging code is still somewhat of a mystery to me too.

I wonder if instead of assigning your webhook instance into the `$api->webhooks` you could instead something like
```
$webhook = new OA\Webhook(...);
$api->merge([$webhook]);
```
instead.

To be honest I also have no idea how Nelmio describers interact with swagger-php, so you might be better off asking over there.

PR #1919, opened 2026-01-15
Fix: Prevent duplicate webhook entries in OpenAPI specification

qdequippe:
### Problem
Without the explicit exclusion of `Webhook` instances, webhooks were being processed twice:

1. **First pass `MergeIntoOpenApi` (line 76)**: `Webhook` annotations were added to the `$merge[]` array.

2. **Second pass `MergeIntoOpenApi` (line 111)**: When `$openapi->merge($merge, true)` is called, the `AbstractAnnotation::merge()` method detects that `Webhook` has an entry in `OpenApi::$_nested` (`Webhook::class => ['webhooks', 'webhook']`) and processes them again through the standard nesting mechanism.

This resulted in duplicate webhook entries in the final OpenAPI specification.

Related to #1917

### DIFF
--- a/src/Processors/MergeIntoOpenApi.php
+++ b/src/Processors/MergeIntoOpenApi.php
@@ -71,7 +71,8 @@ class MergeIntoOpenApi
                 }
             } elseif ($annotation instanceof OA\AbstractAnnotation
                 && in_array(OA\OpenApi::class, $annotation::$_parents)
-                && false === $annotation->_context->is('nested')) {
+                && false === $annotation->_context->is('nested')
+                && false === $annotation instanceof OA\Webhook) {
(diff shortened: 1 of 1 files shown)


DerManoMann:
Seems to work, but I am not sure this is the correct solution.

The fact that the webhook is added in line 76 is due to the fact that its context doesn't have the `nested` flag. In fact, the created `Webhook` doesn't have a context set at all.

Have a look at `Util::createContext` in `NelmioApiDocBundle` which does exactly that...

qdequippe:
thanks @DerManoMann it works perfectly!

```
$context = Util::createContext(['nested' => $api], $api->_context);
$webhook->_context = $context;
```

PR #1944, opened 2026-02-01
Add `NoDynamicNameRule`

DerManoMann:


DerManoMann:
Closing: this patches `phpstan.neon`, which #2142 replaced with `phpstan.neon.dist` (the old name is gitignored now), so the diff no longer applies to anything. The question behind it — whether `NoDynamicNameRule` earns a dev dependency, and what it actually flags at the level we are on — is worth keeping, and it is now tracked with the rest of the phpstan work rather than as a stale draft.

PR #2000, opened 2026-04-22
Fix `parseVarLine()` splitting complex `@var` types at whitespace

### LINKED ISSUE #1999: `@var` generic type fragment incorrectly overrides docblock description
## Description

When a property docblock contains both a description and a `@var` annotation with a generic type (e.g. `array<string, string>`), `DocblockTrait::parseVarLine()` incorrectly splits the generic at the comma-space, treating the trailing part as the description. This causes `AugmentProperties` to set the property description to a type fragment (e.g. `string>`) instead of the actual docblock text.

## Steps to Reproduce

```php
<?php

namespace App;

use OpenApi\Attributes as OA;

#[OA\Schema]
class BugReport
{
    /**
     * An associative array with string values
     * @var array<string, string>
     */
    public array $map;
}
```

```php
$openapi = \OpenApi\Generator::scan(['/path/to/BugReport.php']);
echo $openapi->toJson();
```

## Expected Result

```json
{
  "properties": {
    "map": {
      "description": "An associative array with string values",
      "type": "array",
      "items": { "type": "string" }
    }
  }
}
```

## Actual Result

```json
{
  "properties": {
    "map": {
      "description": "string>",
      "type": "array",
      "items": { "type": "string" }
    }
  }
}
```

krissss:
## Summary

- Fix `DocblockTrait::parseVarLine()` incorrectly splitting generic types (e.g. `array<string, string>`) at internal whitespace, treating the trailing fragment as description
- Replace regex-based type extraction with bracket-depth tracking to correctly handle generic, union, nullable, and array-shorthand types

Closes #1999

### DIFF
--- a/src/Processors/Concerns/DocblockTrait.php
+++ b/src/Processors/Concerns/DocblockTrait.php
@@ -186,14 +186,38 @@ trait DocblockTrait
         $comment = str_replace("\r\n", "\n", (string) $docblock);
         $comment = preg_replace('/\*\/[ \t]*$/', '', $comment); // strip '*/'
 
-        preg_match('/@var\s+(?<type>[^\s]+)([ \t])?(?<description>.+)?+$/im', (string) $comment, $matches);
+        if (!preg_match('/@var\s+(.+)$/im', (string) $comment, $matches)) {
+            return ['type' => null, 'description' => null];
+        }
 
-        $result = array_merge(
-            ['type' => null, 'description' => null],
-            array_filter($matches, static fn ($key): bool => in_array($key, ['type', 'description']), ARRAY_FILTER_USE_KEY)
-        );
+        $rest = $matches[1];
+        $type = '';
+        $depth = 0;
+        $len = strlen($rest);
+        $pos = 0;
 
-        return array_map(static fn (?string $value): ?string => null !== $value ? trim($value) : null, $result);
+        while ($pos < $len) {
+            $char = $rest[$pos];
+            if ('<' === $char || '{' === $char) {
+                ++$depth;
+                $type .= $char;
+            } elseif ('>' === $char || '}' === $char) {
+                --$depth;
+                $type .= $char;
+            } elseif (0 === $depth && ctype_space($char)) {
+                break;
+            } else {
+                $type .= $char;
+            }
+            ++$pos;
+        }
+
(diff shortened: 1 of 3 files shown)


DerManoMann:
Funny that this overlaps with my current PR! I tried to incorporate your changes into #1998 but that seems tricky - however [Claude managed to improve(!) the regexp to cover generics](https://github.com/zircote/swagger-php/pull/1998/changes/2c90851849ef08ddf3a8e363c39cb26f1e1dd479) too 🍤 

Not ideal either as it's really not readable any more. However, I think the long term solution is to use `phpstan/phpdoc-parser` anyway...

DerManoMann:
Actually did just that (refactor to using the parser). I also took the liberty of adding the new test from this branch. Let me know if that is ok - its a very useful test (that I was skirting around)....

This means we can close this PR in favour of #1998 unless I missed anything.

krissss:
Thanks for the heads-up! I did look at #1998 before opening this PR, but at the time it seemed tricky to integrate the bracket-depth-tracking approach alongside your changes, so I went with a separate PR.

Great to hear the test cases are useful — glad they could be folded into #1998.

> the long term solution is to use `phpstan/phpdoc-parser`

I actually tried going down that road before (in another project, also for docblock parsing) but ended up backing out — the overhead of pulling it in for what amounts to a few regex matches felt like overkill. Regex gets a bad rap, but for targeted extractions like `@var` lines it's hard to beat for simplicity and speed.

Closing this in favour of #1998.

krissss:
Closing in favour of #1998.

DerManoMann:
> Great to hear the test cases are useful — glad they could be folded into #1998.
> 
> > the long term solution is to use `phpstan/phpdoc-parser`
> 
> I actually tried going down that road before (in another project, also for docblock parsing) but ended up backing out — the overhead of pulling it in for what amounts to a few regex matches felt like overkill. Regex gets a bad rap, but for targeted extractions like `@var` lines it's hard to beat for simplicity and speed.

Yeah, fair enough. 

Its a bit different here as we have the dependency already anyway and tweaking regexp's that parse (mostly) various docblock fragments is not really what this project is about ;)

DerManoMann:
Thanks @krissss

PR #2003, opened 2026-04-23
Map `array<K, V>` to `type: object` + `additionalProperties`

### LINKED ISSUE #2001: Support `array<K, V>` mapping to `type: object` + `additionalProperties`
## Current Behavior

When a property uses `@var array<K, V>` (e.g. `array<string, string>`), both `LegacyTypeResolver` and `TypeInfoTypeResolver` map it to `type: array` + `items`, treating it as an ordered list.

```php
/**
 * @var array<string, string>
 */
#[OA\Property]
public array $map;
```

Produces:
```json
{ "type": "array", "items": { "type": "string" } }
```

## Expected Behavior

In OpenAPI, `array<K, V>` (two generic parameters) is semantically a map/dictionary, which should be `type: object` + `additionalProperties`. Only `array<T>` (single generic, e.g. `array<string>` or `string[]`) should be `type: array` + `items`.

```json
{ "type": "object", "additionalProperties": { "type": "string" } }
```

## Workaround

Currently requires explicit annotation:

```php
#[OA\Property(
    type: 'object',
    additionalProperties: new OA\AdditionalProperties(type: 'string'),
)]
public array $map;
```

## Notes

- `TypeInfoTypeResolver` uses `phpstan/phpdoc-parser` + `symfony/type-info` which already provides `CollectionType::getCollectionKeyType()` and `CollectionType::getCollectionValueType()`. The key type is available but not used for the `array` vs `object` decision.
- The `CollectionType` branch in `TypeInfoTypeResolver::setSchemaType()` unconditionally sets `$schema->type = 'array'`.
- `LegacyTypeResolver` has the same limitation and is deprecated, so this request primarily targets `TypeInfoTypeResolver`.

krissss:
Closes #2001

## Summary

- `TypeInfoTypeResolver` now maps `array<K, V>` (with explicit key type) to `type: object` + `additionalProperties` instead of `type: array` + `items`
- `array<T>`, `T[]`, `list<T>` (default `int|string` key) keep existing `type: array` + `items` behaviour
- `ArrayShapeType` (e.g. `array{foo:bool}`) also correctly resolves to `type: object` + `additionalProperties`
- `LegacyTypeResolver` is **not** affected — it continues to produce `type: array` + `items` for all array types

## Test plan

- [x] `TypeResolverTest` — 152 tests pass (both 3.0.0 and 3.1.0, both legacy and type-info resolvers)
- [x] `AugmentPropertiesTest` — map assertion merged into `testTypedProperties`
- [x] Nested `array<int, T>` in union types correctly produces `object` + `additionalProperties`


src/Type/TypeInfoTypeResolver.php
```diff
@@ -148,19 +148,37 @@ protected function setSchemaType(OA\Schema $schema, Type $type, Analysis $analys
...
-                if (Generator::isDefault($schema->items)) {
-                    $schema->items = new OA\Items(['_context' => new Context(['generated' => true], $schema->_context)]);
-                    $this->setSchemaType($schema->items, $type->getCollectionValueType(), $analysis);
-                    $this->type2ref($schema->items, $analysis);
-                    $analysis->addAnnotation($schema->items, $schema->items->_context);
-                } elseif (Generator::isDefault($schema->items->type, $schema->items->oneOf, $schema->items->allOf, $schema->items->anyOf)) {
-                    $this->setSchemaType($schema->items, $type->getCollectionValueType(), $analysis);
-                    $this->type2ref($schema->items, $analysis);
-                }
+                if ($type->isList() || $type->getCollectionKeyType() instanceof UnionType) {
```
DerManoMann, line 151: This is tricky. `UnionType` means `int|string` so could be an `object`, we just don't know. However, from what I understand this happens when no key is specified.
I have to admin I do not quite like this check as it is quite loose, however I do not know if there is a better alternative... hmmm.

DerManoMann (commented):
Looks like it is doing what it should. I do not quite like the `UnionType` check, but there might be no alternative.

Only think to consider, even thought there seems to be a bit of variation, is wheter there is now a lot of duplication that could be simplified around the creation of nested `Schema`.

DerManoMann:
Tried to come up with an example that might conflict with the union check over the weekend...
How about this:

```php
enum TypeEnum: string
{
    case First = 'first';
    case Second = 'second';
}

#[OAT\Schema()]
class TypedProperties
{
    /**
     * A map of stringy to string.
     *
     * @var \ArrayObject<TypeEnum|string,string> $stringyMap
     */
    #[OAT\Property]
    public \ArrayObject $stringyMap;
}
```

DerManoMann:
Claude suggests to check if all union types are built in types, but that still feels ambiguous to me. What about checking if the union contains `int`?

Casting rules around [array keys are tricky](https://phpstan.org/blog/why-array-string-keys-are-not-type-safe) so maybe requiring to be explicit about `int` is a good thing?

krissss:
Thanks for the detailed feedback and the edge case example!

**Regarding the `array<int, string>` case:**

In PHP, `json_encode` only produces a JSON array when keys are sequential starting from `0`. For non-sequential int keys like `[1 => 'a', 3 => 'b']`, the result is a JSON object `{"1":"a","3":"b"}`.

The key insight is: if someone writes `array<int, string>`, they are explicitly declaring a key type — which means they intend a map. If they wanted a sequential list, they would simply write `array<string>`, `string[]`, or `list<string>`. The explicit key type is what distinguishes the two.

**Regarding the `ArrayObject<TypeEnum|string, string>` example:**

PHP array keys only allow `int` or `string`, so `TypeEnum` would be cast to its string backing value at runtime. While the PHPDoc annotation can declare such a key type, this pattern is extremely uncommon in practice. The vast majority of real-world cases are `array<string, V>` and `array<int, V>`.

**Regarding the "check if key contains `int`" suggestion:**

This would cause `array<int, string>` to be treated as `type: array`, but as discussed above, anyone writing `array<int, string>` most likely intends a map, not a list.

DerManoMann:
Yeah, I agree this is a bit of a mess. Also agree that my example is a bit far fetched ;)

Let's go with this for now. We can always refine this if needed.

DerManoMann (approved):
Thanks @krissss

PR #2008, opened 2026-04-29
Make `AbstractAnnotation::$_context` non-nullable

### LINKED ISSUE #2007: AbstractAnnotation::$_context is typed as nullable but can never be null
## Description

`AbstractAnnotation::$_context` is declared as `public ?Context $_context`, but the constructor guarantees it is always initialized to a non-null `Context` instance:

```php
public function __construct(array $properties)
{
    if (isset($properties['_context'])) {
        $this->_context = $properties['_context'];
    } elseif (Generator::$context) {
        $this->_context = Generator::$context;
    } else {
        $this->_context = new Context(['generated' => true]);
    }
    // ...
}
```

All three branches assign a non-null `Context`. This was introduced in #1947 when converting from PHPDoc `@var Context|null` to native type `?Context`. While the original PHPDoc was `Context|null` (likely because the untyped property had no default value in older PHP versions), the native type should now reflect the actual runtime behavior.

## Motivation

PHPStan (level 6+) reports `$_context` as nullable, leading to unnecessary null checks in downstream code or requiring stub files to patch the type. Making the type non-nullable would improve static analysis accuracy for all users.

krissss:
Closes #2007

## Summary

- Change `AbstractAnnotation::$_context` type from `?Context` to `Context`
- The constructor guarantees a non-null `Context` across all three initialization branches
- The nullable type was a leftover from the pre-native-type PHPDoc `@var Context|null`, converted in #1947

## Test plan

- [x] Full test suite passes (1037 tests)
- [ ] CI matrix

### DIFF
--- a/src/Annotations/AbstractAnnotation.php
+++ b/src/Annotations/AbstractAnnotation.php
@@ -35,7 +35,7 @@ abstract class AbstractAnnotation implements \JsonSerializable
      */
     public $attachables = Generator::UNDEFINED;
 
-    public ?Context $_context;
+    public Context $_context;
 
     /**
      * Annotations that couldn't be merged by mapping or postprocessing.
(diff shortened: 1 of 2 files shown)


DerManoMann (approved):

DerManoMann:
Thanks @krissss

PR #2011, opened 2026-04-30
fix: prevent PHP "mixed" from being emitted as invalid OpenAPI "type: mixed"

### LINKED ISSUE #2010: PHP "mixed" type emitted as invalid OpenAPI "type: mixed"
## Summary

swagger-php can emit schemas containing `type: mixed`, which is not a valid OpenAPI Schema Object type.

This occurs for PHPDoc types such as `array<string, mixed>` where the generated schema may contain:

```yaml
type: object
additionalProperties:
  type: mixed
```

or for array items inferred as `mixed`.

## Why this is incorrect

OpenAPI 3.0 and 3.1 follow JSON Schema types, whose valid `type` values are: `null`, `boolean`, `object`, `array`, `number`, `string`, `integer`.

`mixed` is not among them.

- OpenAPI 3.1.1 Data Types: https://spec.openapis.org/oas/v3.1.1.html#data-types
- OpenAPI 3.0.0 Data Types: https://spec.openapis.org/oas/v3.0.0.html#data-types

## Root cause

`TypeResolverInterface::NATIVE_TYPE_MAP` (`src/TypeResolverInterface.php:14`) contains:

```php
'mixed' => 'mixed',
```

`AbstractTypeResolver::mapNativeType()` (`src/Type/AbstractTypeResolver.php:29-58`) treats `mixed` as a recognized native type and writes it back to the OpenAPI schema as `type: mixed`.

Additionally, `LegacyTypeResolver.php:164` directly assigns `$types = ['mixed']` when the PHP native type is bare `array` (without docblock generics).

## Expected behavior

PHP `mixed` should be represented as an unconstrained schema (empty `{}`) rather than a `type` value:

```yaml
# mixed → unconstrained
{}

# mixed[] → array of anything
type: array
items: {}

# array<string, mixed> → object with unconstrained additionalProperties
type: object
additionalProperties: {}
```

## Minimal example

```php
use OpenApi\Attributes as OA;

#[OA\Schema]
final class ExampleDto
{
    /** @var array<string, mixed> */
    #[OA\Property]
    public array $extra = [];
}
```

**Expected output:**
```yaml
extra:
  type: object
  additionalProperties: {}
```

**Actual output:**
```yaml
extra:
  type: object
  additionalProperties:
    type: mixed
```

## Affected code paths

| Scenario | LegacyTypeResolver | TypeInfoTypeResolver |
|----------|-------------------|---------------------|
| `array<string, string>` | `items: { type: mixed }` | `additionalProperties: { type: string }` (correct) |
| `array<int, string>` | `items: { type: mixed }` | `additionalProperties: { type: string }` (correct) |
| `mixed $prop` (no docblock) | `type: mixed` | `type: mixed` |
| `string\|array` mixedUnion | correctly omits type | `items: { type: mixed }` |

Note: the test suite (`tests/Type/TypeResolverTest.php:48-50, 97-99`) currently asserts `type: mixed` as expected output, so tests pass but produce invalid OpenAPI.

krissss:
## Summary

Fixes #2010

PHP `mixed` type was being emitted as `type: mixed` in OpenAPI schemas, which is invalid — OpenAPI only supports: `null`, `boolean`, `object`, `array`, `number`, `string`, `integer`.

### Before
```yaml
extra:
  type: object
  additionalProperties:
    type: mixed    # invalid OpenAPI
```

### After
```yaml
extra:
  type: object
  additionalProperties: {}    # unconstrained — matches PHP mixed semantics
```

## Changes

- **`AbstractTypeResolver::mapNativeType()`** — when the resolved type is `mixed`, return `true` (recognized) without setting `schema->type`, producing an unconstrained schema
- **`LegacyTypeResolver::augmentItems()`** — when parent schema type is `mixed`, items use unconstrained type instead of inheriting `mixed`; added guard for items existence
- **`TypeInfoTypeResolver::setSchemaType()`** — skip `BuiltinType::MIXED`, don't write it to schema
- **Test expectations** — updated 6 assertions from `type: mixed` to unconstrained (`{}`)

## Design decision

`NATIVE_TYPE_MAP` retains `'mixed' => 'mixed'` so that downstream nullable logic continues to work correctly. The interception happens at the point where the type would be written to the OpenAPI schema — `mixed` is recognized as a valid PHP type but mapped to an unconstrained schema (no `type` set) rather than the invalid `type: mixed`.


DerManoMann (approved):
Nice one!

PR #2028, opened 2026-06-23
fix: drop unmappable types in nested and union schemas

alleknalle:
swagger-php only dropped types with no OpenAPI representation (callable, resource, ...) at the top level, so nested and union positions emitted an invalid type (e.g. items: {type: callable}). Drop them in setSchemaType, leaving the schema open instead.

## Summary

swagger-php only applies its "drop unmappable type" sanity check to the top-level schema, so nested and union positions leak the raw PHP type name and emit invalid output:

| Input | Before | After |
|---|---|---|
| `list<callable>` | `items: {type: callable}` | `items: {}` |
| `array<string, callable>` | `additionalProperties: {type: callable}` | `additionalProperties: {}` |
| `int\|callable` | `type: [callable, integer]` | `type: [integer]` |

The fix drops these in `setSchemaType`, where the type is known to be a `BuiltinType` (callable, resource, iterable, …) rather than a class — so object/class types still resolve to `$ref`s as before. `mapNativeType` id: it must keep passing class-name strings through for later `$ref` resolution.

Out of scope: the rare non-all-builtin union where the builtin member itself is unmappable (e.g. `Foo|callable`); happy to extend if wanted.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Related Issues

None.

## Test Plan
Resolver-level cases in `tests/Type/TypeResolverTest` (`list<callable>`, `array<string, callable>`) and an end-to-end case in  tests/Processors/AugmentPropertiesTest` (via the `AugmentProperties` processor). `composer test` + `composer analyse` green.

- [x] Unit tests added/updated
- [x] Integration tests added/updated

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code where necessary
- [x] I have updated the documentation accordingly 
- [x] My changes generate no new warnings
- [x] New and existing tests pass locally
- [x] Any dependent changes have been merged and published


src/Type/TypeInfoTypeResolver.php
```diff
@@ -140,7 +144,8 @@ protected function setSchemaType(OA\Schema $schema, Type $type, Analysis $analys
             }
         } else {
             if ($type instanceof BuiltinType) {
-                if (!$type->isIdentifiedBy(TypeIdentifier::MIXED)) {
+                // Only set a type that has an OpenAPI representation; mixed, callable, resource, iterable etc. stay open instead of emitting an invalid type.
```
DerManoMann: could this comment be moved (and adjusted) to the method docblock in stead, please?
`Checks that the given type ...`

alleknalle: Done

DerManoMann:
Thanks @alleknalle

PR #2034, opened 2026-06-30
feat: populate schema required from its properties

### LINKED ISSUE #2017: [Feature]: Automatically populate schema::required from properties that are `required`
### Problem Statement

Just a thought

### Proposed Solution

Automatically generate `Schema::required` based on `Property::required` values

### Alternatives Considered

_No response_

### Priority

priority/critical

### Area

None

### Additional Context

_No response_

alleknalle:
Adds an AugmentRequired processor: explicit `required: true`/`false` always applies, and nullability inference is opt-in via `augmentRequired.enabled`.

Closes #2017

## Summary

Today `Schema::required` (the array of required property names) has to be maintained separately from the properties it refers to. This adds a new `AugmentRequired` processor that derives it from the properties themselves, in two layers:

**Explicit per-property flag (always on).** A property can declare `required: true` or `required: false`. `true` adds its name to the parent `Schema::required`, `false` removes it. The boolean is consumed during processing so it never serialises on the property, where `required` must be an array.

```php
#[OA\Property(property: 'name', type: 'string', required: true)]
```

**Nullability inference (opt-in).** With the `augmentRequired.enabled` config flag (off by default, same shape as `cleanUnusedComponents.enabled`), a property backed by a PHP member becomes required when it has a known, non-nullable type:

```php
#[OA\Property] public string $name;       // required
#[OA\Property] public ?string $nickname;  // optional
```

### Design notes

- **Explicit wins, inference fills the gaps.** An explicit `required`/`nullable` value always takes precedence over inference, and an explicitly declared `required` list on a schema is never modified by inference.
- **Inference only marks a positively-known he `!isNullable()` rule used for parametersand request bodies, property scanning also hits untyped/`mixed`/unresolvable properties; those have no determinable nullability, so they are left optional rather than assumed required.
- **Inference is scoped to PHP-backed memberomoted parameter, method), so hand-authoredinline schema properties are untouched.
- **Off by default** because inference changes output for existing projects that use typed properties; the explicit flag is new syntax and therefore safe to always run.

### API change (non-breaking)

The `Property` attribute's `required` parameter widens from `?array` to `bool|array|null`. Arrays and `null` behave exactly as before; only the boolean form is new.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Enhancement (improvement to existing functionality)
- [ ] Breaking change (fix or feature that wlity to change)
- [x] Documentation update
- [ ] Refactor (no functional changes)
- [ ] Chore (maintenance, dependencies, CI)

## Related Issues

Closes #2017

## Test Plan

**Unit / processor tests** — `AugmentRequiredTest` (16 cases) plus fixtures cover: the opt-in flag, non-nullable typed properties (incl. `$ref`), nullable/untyped/`mixed` left optional, explicit `nullable` override, promoted-parameter and
method-backed properties, the reflector gate, explicit `true`/`false` adding to andremoving from both empty and declared lists, the boolean being consumed (named and unnamed properties), and an array `required` on an object property left untouched. The `InvalidPropertyAttribute` test now passes a string to still
assert the constructor's type error.

**Integration test** — a `ScratchTest` fixture (`RequiredFromProperties`) generates a full spec for a schema with explicit `required: true`/`false` properties referenced from a `GET` endpoint, and compares the serialized YAML across OpenAPI 3.0/3.1/3.2 and both type resolvers. This verifies end-to-end that `required: true` emits `required: [id]` on the component and that no `required: true` leaks onto a property in the output.

Ran locally: full suite (1081 tests), PHPStan, php-cs-fixer, and `rector --dry-run` all pass.

- [x] Unit tests added/updated
- [x] Integration tests added/updated
- [x] Manual testing performed

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code where necessary
- [x] I have updated the documentation accordingly
- [x] My changes generate no new warnings
- [x] New and existing tests pass locally
- [x] Any dependent changes have been merged and published

### DIFF
--- a/docs/reference/annotations.md
+++ b/docs/reference/annotations.md
@@ -853,6 +853,10 @@ A <code>@OA\Request</code> path parameter.
 <dl>
   <dt><strong>property</strong> : <span style="font-family: monospace;">string</span></dt>
   <dd><p>The key into Schema->properties array.</p><table class="table-plain"><tbody><tr><td><i>Required</i>:</td><td style="padding-left: 0;"><b>no</b></td></tr></tbody></table></dd>
+  <dt><strong>required</strong> : <span style="font-family: monospace;">bool|list&lt;string&gt;</span></dt>
+  <dd><p>A boolean flags the property as required in its parent schema, or a list names the required members when it is an object.<br />
+<br />
+The <code>AugmentRequired</code> processor collects a boolean into the parent <code>Schema::$required</code>.</p><table class="table-plain"><tbody><tr><td><i>Required</i>:</td><td style="padding-left: 0;"><b>no</b></td></tr></tbody></table></dd>
 </dl>
 
 ### [Put](https://github.com/zircote/swagger-php/tree/master/src/Annotations/Put.php)
--- a/docs/reference/attributes.md
+++ b/docs/reference/attributes.md
@@ -2238,9 +2238,10 @@ value of this attribute.</p><table class="table-plain"><tbody><tr><td><i>Require
   <dd><p>The minimum number of properties allowed in an object instance.<br />
 An object instance is valid against this property if its number of properties is greater than, or equal to, the<br />
 value of this attribute.</p><table class="table-plain"><tbody><tr><td><i>Required</i>:</td><td style="padding-left: 0;"><b>no</b></td></tr></tbody></table></dd>
-  <dt><strong>required</strong> : <span style="font-family: monospace;">list&lt;string&gt;</span></dt>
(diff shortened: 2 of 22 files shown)


DerManoMann:
closing for now - perhaps worth doing for the new spec pipeline. I'll leave #2017 open so we do not forget.

PR #2077, opened 2026-07-20
feat(BETA): add `AttributeTranslatorInterface` concept

DerManoMann:
## Summary

Introduces a new interface `AttributeTranslatorInterface` and toolchain related methods.

This interface hooks deep into the `AttributeFactory` and allows custom implementations to be added. The contract includes two methods:

- `getAttributes()` - return a list of `ReflectionAttribute` instances to be instatiated for the given `\Reflector`. These can be `OpenApi` attributes or unrelated
- `translate()` - expected to convert the loaded attributes into `AttibuteInterface` instances. This allows to map custom/framework attributes directly into raw OpenApi attributes. If they are in the right place in the code structure, nothing else needs to be done and the `Assembler` should merge/nest them where they belong.

Apart from allowing to inject new attributes, this can also be used as low level customization. For example, a validation constrait attribute could be used to set some value on a `Schema` instance, etc.

### DIFF
--- a/src/Utils/AttributeFactory.php
+++ b/src/Utils/AttributeFactory.php
@@ -8,6 +8,7 @@ namespace OpenApi\Utils;
 
 use OpenApi\Assembler\DefaultAttributeTranslator;
 use OpenApi\AttributeInterface;
+use OpenApi\AttributeTranslatorInterface;
 use OpenApi\OpenApiException;
 use OpenApi\Spec as OA;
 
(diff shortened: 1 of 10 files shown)


src/Utils/AttributeFactory.php
```diff
@@ -275,32 +319,29 @@ protected function nestChild(AttributeInterface $parent, AttributeInterface $chi
...
-            \ReflectionAttribute::IS_INSTANCEOF,
-        );
-
-        $result = [];
-        foreach ($attributes as $attribute) {
-            if (!class_exists($attribute->getName())) {
-                continue;
-            }
+        $attributes = [];
+        foreach ($this->translators as $translator) {
```
DerManoMann, line 322: Still need to split off internal translator(s) from the public/optional list.

PR #2150, opened 2026-09-01
refactor(Spec): use `Undefined::UNDEFINED` for every mixed property

DerManoMann:
## Overview

A `mixed` property cannot use `null` to mean "not set", because `null` is a legal value for
it — a schema default, an example or a link request body may genuinely be null. Twelve of
the fourteen `mixed` properties in `src/Spec/` already defaulted to `Undefined::UNDEFINED`
for that reason, but the convention was written down nowhere and two properties deviated
from it.

This aligns the two deviations, writes the rule down, and adds a test that keeps it true.
Compiled output is unchanged: `filter()` drops `null` and `Undefined::UNDEFINED` alike, so
the only behavioural difference is that an explicit `null` can now be expressed and emitted.

Split out of #2149 ahead of the OpenAPI 3.2 field work, which it does not depend on.

## Changes

- `Example::$value` and `Link::$requestBody` default to `Undefined::UNDEFINED`
- `OpenApi31Compiler` emits both explicitly rather than through `filter()`, matching what
  `compileSchema()` already did for `default`, `const` and `example`
- `HybridBridge` passes both straight through — classic's `Generator::UNDEFINED` is an alias
  of the same sentinel, so no translation is needed
- `UndefinedDefaultsTest` asserts every `mixed` property in `src/Spec/` follows the convention
- `docs/dev/pipeline.md` gains "`null` means unset, except where `null` is a value"
- drops a redundant inline `@var` on `OpenApi::$security`, duplicated from the constructor's
  `@param`


src/Compiler/OpenApi31Compiler.php
```diff
@@ -396,14 +396,19 @@ protected function compileLink(OA\Link $link): array
...
+        $result = $this->filter([
             'operationRef' => $link->operationRef,
             'operationId' => $link->operationId,
             'parameters' => $link->parameters,
-            'requestBody' => $link->requestBody,
             'description' => $link->description,
             'server' => $link->server instanceof OA\Server ? $this->compileServer($link->server) : null,
         ], $link);
+
+        if ($link->requestBody !== Undefined::UNDEFINED) {
```
DerManoMann, line 407: is this gating necessary if `filter` will strip it anyway (just like `null`)?

src/Compiler/OpenApi31Compiler.php
```diff
@@ -641,12 +646,17 @@ protected function compileFlow(OA\Flow $flow): array
...
     {
-        return $this->filter([
+        $result = $this->filter([
             'summary' => $example->summary,
             'description' => $example->description,
-            'value' => $example->value,
             'externalValue' => $example->externalValue,
         ], $example);
+
+        if ($example->value !== Undefined::UNDEFINED) {
```
DerManoMann, line 655: Same here?

src/Compiler/OpenApi31Compiler.php
DerManoMann: It is necessary, but not to strip `UNDEFINED` — it is there so an explicit `null` survives. `filter()` strips all three values:

```php
$result = array_filter($result, fn ($value): bool => !in_array($value, [null, Undefined::UNDEFINED, []], true));
```

So with `requestBody` inside `filter()`, `requestBody: null` is indistinguishable from unset and gets dropped — which is the defect this fixes, since `null` is a legal value. Assigning after `filter()` is the only thing that makes "set to null" differ from "not set". The same applies to `[]`, which `filter()` also strips.

Putting it back inside `filter()` fails `testLinkRequestBodyNullIsEmitted` while `testLinkRequestBodyUnsetIsOmitted` still passes, so the two tests pin both directions.

This is also the pattern `compileSchema()` already uses for `default`, `const` and `example` — the change makes the five `mixed` properties consistent rather than introducing anything new.

Added a note to `pipeline.md` that `filter()` drops `[]` as well, since that is the part which makes this look redundant.

src/Compiler/OpenApi31Compiler.php
DerManoMann: Same reason as above: `filter()` strips `null`, `Undefined::UNDEFINED` and `[]`, so leaving `value` inside it drops an explicit `null` — and `null` is a legal example value. Verified by reverting it: `testExampleValueNullIsEmitted` fails with "Failed asserting that an array has the key 'value'", `testExampleValueUnsetIsOmitted` still passes.

PR #2178, opened 2026-09-09
fix(Annotations): stop treating a schema's examples as Example Objects

### LINKED ISSUE #2177: [Bug]: @OA\Examples used under Schema emits spurious validation errors for summary and example key-field
### Description

  Since PR #2176 fixed schema-level `examples` serialisation, `@OA\Examples` can be nested under `@OA\Schema` / `@OA\Property` / `@OA\JsonContent` to  produce a JSON Schema–compliant `"examples": [value]` array. However, the classic pipeline's validator still treats every `@OA\Examples` instance as if it were in a media-type / parameter context, which causes two spurious diagnostic messages that cannot be silenced without adding meaningless boilerplate: 
  
  ### 1. Missing key-field error
  
  ```
  [error] @OA\Examples() is missing key-field: "example" in … on line N
  ```
  
  `$_nested` on `Schema` declares `Examples::class => ['examples', 'example']`, making `example` the **map key**. But `exampleValues()` deliberately ignores the key and extracts only `value` — the key is irrelevant for a JSON Schema list. The validator still errors if `example` is absent.
  
  ### 2. Missing required field warning
  
  ```
  [warning] Missing required field "summary" for @OA\Examples() in … on line N
  ```
  
  `Examples::$_required = ['summary']` is inherited unconditionally. `summary` is meaningful only in an Example Object (media-type / parameter context);
  `exampleValues()` does not use it and it does not appear in the serialised output.

  ## Current workaround
  
  Users must supply dummy values that are silently discarded:
  
  ```php
  examples: [new OA\Examples(example: "default", summary: "", value: 22)]
  ```
  ## Suggested fix
  
  Override `validate()` in `Schema` (or add a context check in `Examples::validate()`) to suppress the key-field and `$_required` checks when the parent is
  a Schema. Alternatively, introduce a separate `@OA\SchemaExample` annotation (or a plain-value shorthand) for the JSON Schema list context — removing the
  ambiguity entirely.

### Steps to Reproduce

  ```php
  #[OA\Schema]
  class Foo {
      #[OA\Property(
          type: "integer",
          examples: [new OA\Examples(value: 22)]  // no `example` key, no `summary`
      )]  
      public int $port;
  }   
  ``` 
  
  Running `(new Generator())->generate([...])` produces:
  
  ```
  [error]   @OA\Examples() is missing key-field: "example" in Foo->port
  [warning] Missing required field "summary" for @OA\Examples() in Foo->port
  ```
  
  Even though the generated spec is correct: `"examples": [22]`.

### Expected Behavior

When `@OA\Examples` is nested under a Schema, neither the `example` key-field nor `summary` should be required. Validation should be context-aware — skip those checks when the parent is a Schema (or its subclasses).

### Actual Behavior

When `@OA\Examples` is nested under a Schema/Property, the generator emits
  two diagnostics even though the serialised output is correct:

  [error]   @OA\Examples() is missing key-field: "example"
  [warning] Missing required field "summary" for @OA\Examples()

  To suppress them, users are forced to supply dummy values that are silently
  discarded by `exampleValues()` and never appear in the generated spec:
  
      examples: [new OA\Examples(example: "default", summary: "", value: 22)]

### Priority

priority/medium

### Area

None

### Environment

Version: 6.8.0

### Logs or Screenshots

_No response_

DerManoMann:
Closes #2177.

## Overview

6.8.0 made a schema's `examples` a list of values, as JSON Schema requires, but left the
nesting declaration saying they were Example Objects. Anyone writing `@OA\Examples` under an
`@OA\Schema` or `@OA\Property` got two diagnostics for a document that generated correctly —
a missing `example` key-field and a missing `summary` — and could only silence them by
supplying values that are then discarded.

The reporter also asked for the plain JSON Schema spelling. It did not work: classic dropped
bare values silently, so an Example Object wrapper was the only way to write a list of
literals, which is what led into the diagnostics in the first place. Both spellings work now,
and classic accepts what the spec pipeline already documented.

## Changes

- `@OA\Examples` nested under a schema no longer asks for an `example` key or a `summary`;
  neither reaches the output. Media types, parameters and headers are untouched — theirs are
  Example Objects and keep every field.
- A schema's `examples` accepts plain values, in attributes and docblocks alike, and preserves
  the order they were written in when mixed with `@OA\Examples`.
- An example that carries no `value` is now reported instead of vanishing from the list, as is
  one setting both `value` and `externalValue`, and any other annotation passed as an example.
- `Items` and `AdditionalProperties` inherit their nesting from `Schema` rather than holding
  copies of it. `JsonContent` and `XmlContent` declare their own `examples`, which stay Example
  Objects for the generated media type.
- `identity()` no longer assumes a nested entry has a key, which had made every `@OA\Examples`
  print without its name.
- New cookbook section for schema examples. The `-spec` scratch expectations #2175 added for
  3.1 and 3.2 are removed, so all three modes are held to one expectation per version.


DerManoMann:
Merged in 08c8a2d2, via #2179.

#2179 was stacked on this branch but based on `master`, since a cross-repo pull request cannot target a branch on the fork. Its diff therefore carried all six commits, and squash-merging it landed both changes under #2179's title. Nothing is left here — `git diff master fix/schema-examples-validation` shows only this branch lacking the `Items` docblock fix.

Closing as merged. #2177 is fixed by that commit.

Issue #2180, opened 2026-09-09
symfony/console ^7.4 requirement locks the library out of Symfony 6.4 LTS, though only the CLI needs it


garak:
### Problem Statement

Every release from 6.1.2 to 6.8.0 carries the same constraint.

### Why this is worth revisiting

**Symfony 6.4 is an LTS**, in security support until November 2027, with a large installed base.
Applications on it are not misconfigured or behind, they are on a supported release.

**The constraint is inconsistent with the other Symfony dependencies.** From 6.8.1's
composer.json:

| package | constraint |
| --- | --- |
| symfony/console | `^7.4 \|\| ^8.0` |
| symfony/finder | `^5.0 \|\| ^6.0 \|\| ^7.0 \|\| ^8.0` |
| symfony/yaml | `^5.4 \|\| ^6.0 \|\| ^7.0 \|\| ^8.0` |
| symfony/deprecation-contracts | `^2 \|\| ^3` |

Finder goes back to 5.0 and Yaml to 5.4. Console alone demands the newest LTS.

**The library core does not use symfony/console.** `Symfony\Component\Console` appears in three
files only:

- `src/Console/GenerateCommand.php`
- `src/Console/GenerateInput.php`
- `bin/openapi`

`Generator`, the analysers, the annotations and the processors never touch it. Most consumers
reach this library through a framework integration and call `Generator` directly, never
`bin/openapi`. `nelmio/api-doc-bundle`, for instance, requires `^5.7.8 || ^6.0` and uses the
generator, not the command. Those users pay for a CLI dependency they do not run.

**Being stuck on 6.0.6 means shipping an invalid spec.** Concrete case: in 6.0.6
`Annotations\Examples::$_required = ['summary']`, so a Schema Object written per OAS 3.1 (a JSON
Schema 2020-12 dialect, where `examples` MUST be an array) triggers
`Missing required field "summary" for @OA\Examples(example="0")`. 6.8.1 fixed exactly this with
`isSchemaExample()`, which requires only `value` for the schema keyword. On Symfony 6.4 that fix
is out of reach, so the options are a doc page that errors, or rewriting `examples` as a
named-example map, which is invalid inside a Schema Object and rejected by other validators
(apidom reports it as error 10067 `examples must be an array`). Neither is acceptable.

### Proposed Solution

Keep the 7.4 requirement where the code that needs it lives. The floor comes from the CLI
[...]

PR #2196, opened 2026-09-12
Drop symfony/console 7.4 requirement, restore attribute-free command def

yceruto:
## Problem

The library required `symfony/console` ^7.4, even though only the CLI (`bin/openapi`) uses it. This locked out Symfony 6.4 LTS users, who are on a supported release, not an outdated one. It also blocked fixes like the schema `examples` array fix, which only landed in Console 6.8.1.

## Solution

This PR lowers the `symfony/console` constraint and rewrites `GenerateCommand` and `GenerateInput` to use `configure()` and `InputInterface` instead of the `#[MapInput]`, `#[Argument]`, and `#[Option]` attributes, which required Console 7.3+. This keeps one code path for the CLI while letting library consumers depend on older, still-supported Console versions.

Refs #2180

### DIFF
--- a/bin/openapi
+++ b/bin/openapi
@@ -41,6 +41,6 @@ $options = $definition->getOptions();
 unset($options['version'], $options['no-interaction']);
 $definition->setOptions($options);
 
-$app->addCommand(new GenerateCommand($logger));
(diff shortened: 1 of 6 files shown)


DerManoMann:
Thanks @yceruto

PR #2197, opened 2026-09-13
fix(Console): support symfony/console 5.4 and up

### LINKED ISSUE #2180: symfony/console ^7.4 requirement locks the library out of Symfony 6.4 LTS, though only the CLI needs it

garak:
## Overview

The `symfony/console` requirement of `^7.4` kept swagger-php off Symfony 6.4 LTS, which is supported until November 2027. The console component is only used by the CLI, and the requirement came from the command layer using the input-mapping attributes that only exist as of 7.4.

The CLI surface is unchanged: `openapi -h` reports the same arguments and options, and `--version` still sets the OpenAPI version rather than printing the tool
version.

Closes #2180.

## Changes

- `symfony/console` is accepted from 5.4 (generally, the `symfony/*` constraints are now consistent)
- `GenerateCommand` is a `Command` subclass again, defining its input in `configure()` and running in `execute()`.
- `GenerateInput` is a plain DTO, hydrated by `GenerateInput::fromInput()`, which also validates `--format` and `--mode`.
- `bin/openapi` registers the command through whichever of `Application::addCommand()` and `Application::add()` the installed console offers, and its `ArgvInput` subclass no longer types `hasParameterOption()`.
- `DocsAccuracyTest` compares the command's own help only; the built-in options below `--help` differ per console version.


garak:
Superseded by #2916

PR #2202, opened 2026-09-15
fix(Compiler): don't drop an explicit example: [] as if it were unset

DerManoMann:
## Overview

`Compiler::filter()` strips `null`, the `Undefined` sentinel, and `[]` from every compiled
field alike. That's the right call for the derived maps it shares the array with —
`examples`, `encoding`, `headers` and `links` all come back as `[]` from `compileKeyedMap()`
when nothing was declared, and an absent map should disappear from the output. `example`
isn't one of those: its unset state is the `Undefined` sentinel, not `[]` (see
`Spec\MediaType`/`Spec\Parameter`/`Spec\Header`), so a caller-declared `example: []` — a
real, meaningful value ("the example is an empty list") — was indistinguishable from not
declaring one at all, and got silently dropped by the spec/hybrid pipeline while classic
mode preserved it correctly.

Found generating a real project (phpMyFAQ) under hybrid mode: six responses across
`FaqController`, `SearchController` and `PdfController` declare
`content: new OA\JsonContent(example: [])` on empty-result responses, and hybrid mode
dropped `example` from all six where classic kept it.

## Changes

- Added `Compiler::withDefined()`, which gates a key on the `Undefined` sentinel directly
  instead of routing it through `filter()`.
- Used it in `compileMediaType()`, `compileParameter()` and `compileHeader()` for `example`,
  fixing the bug.
- Also used it in `compileExample()` (`value`), `compileLink()` (`requestBody`) and
  `compileSchema()` (`default`/`const`/`example`), which already hand-rolled the identical
  sentinel check inline — same fix, same shape, one helper instead of six copies.
- Added `CompilerTest::testExampleEmptyArrayIsEmitted()` /
  `testExampleUnsetIsOmitted()`, each data-provider-driven across MediaType/Parameter/Header.


src/Compiler/OpenApi31Compiler.php
```diff
@@ -603,15 +602,9 @@ protected function compileSchema(OA\Schema|string $schema): array|\stdClass
...
-        if ($schema->default !== Undefined::UNDEFINED) {
-            $result['default'] = $schema->default;
-        }
-        if ($schema->const !== Undefined::UNDEFINED) {
-            $result['const'] = $schema->const;
-        }
-        if ($schema->example !== Undefined::UNDEFINED) {
-            $result['example'] = $schema->example;
-        }
+        $result = $this->withDefined($result, 'default', $schema->default);
```
DerManoMann: Could we allow multiple into the method? 2nd parameter as `[]`?

DerManoMann: Done in 7574e185 — `withDefined(array $result, array $values)` now, so `compileSchema()` is one call:

```php
$result = $this->withDefined($result, [
    'default' => $schema->default,
    'const' => $schema->const,
    'example' => $schema->example,
]);
```

The five single-field call sites pay an array literal each (`withDefined($result, ['example' => $parameter->example])`), which seemed the better trade against carrying a `string|array` second parameter.

Note `OpenApi30Compiler::compileSchema()` still hand-rolls the sentinel checks — `example` there has an `elseif` fallback to `examples[0]` and `const` only maps to `enum` when `enum` isn't already set, so only `default` would convert cleanly. Left as is.

